PR #1, opened 2025-12-10
feat: HLL backend implementation

### LINKED ISSUE #3: Implement HLL
Using https://github.com/apache/datasketches-cpp/tree/master/hll as a reference, we should aim specifically to have a datastructure that can serialize / deserialize the C++ / Java / Go implementation.

notfilippo:
Closes #3 

Currently adds support for: `List`, `HashSet`, `Array4`, `Array6`, `Array8` with serialisation / deserialisation support. 

The format is compatible with other datasketch implementations and it is tested by deserialising and re-serialising sketches. The test data is [borrowed from the datasketches-go library](https://github.com/apache/datasketches-go/tree/main/serialization_test_data). 

What's left to do:
- ~~Implement methods on the high-level sketch wrapper `HllSketch` (creation, promotion)~~
- ~~Improve test coverage, currently most of the focus on testing went in the serialisation check but some unit test could be improved and expanded~~
- ~~Finalise public API~~

### DIFF
--- a/src/hll/aux_map.rs
+++ b/src/hll/aux_map.rs
@@ -97,10 +97,6 @@ impl AuxMap {
     }
 
     /// Insert a new slot-value pair
-    ///
-    /// # Panics
-    ///
-    /// If the slot already exists in the map
     pub fn insert(&mut self, slot: u32, value: u8) {
         let index = self.find(slot);
         match index {
@@ -128,10 +124,6 @@ impl AuxMap {
     }
 
     /// Replace value for existing slot
-    ///
-    /// # Panics
-    ///
-    /// If the slot doesn't exist in the map
     pub fn replace(&mut self, slot: u32, value: u8) {
         match self.find(slot) {
             FindResult::Found(idx) => {
--- a/src/hll/sketch.rs
+++ b/src/hll/sketch.rs
@@ -71,7 +71,7 @@ impl HllSketch {
     ///
     /// # Panics
     ///
-    /// Panics if lg_config_k is not in [4, 21]
+    /// If lg_config_k is not in range [4, 21]
     pub fn new(lg_config_k: u8, hll_type: HllType) -> Self {
         assert!(
             lg_config_k > 4 && lg_config_k < 21,
--- a/src/hll/composite_interpolation.rs
+++ b/src/hll/composite_interpolation.rs
@@ -55,7 +55,7 @@ pub const fn get_x_arr_length() -> usize {
 ///
 /// # Panics
 ///
-/// If lg_k is not supported (currently supports 4-12, full support requires 4-21)
+/// If lg_k is not in range [4, 21]
 pub fn get_x_arr(lg_k: u8) -> &'static [f64; NUM_X_VALUES] {
     if !(4..=21).contains(&lg_k) {
         panic!("lg_k must be in range [4, 21], got: {}", lg_k);


src/hll/mod.rs
```diff
@@ -0,0 +1,115 @@
...
+pub mod composite_interpolation;
+pub mod container;
+pub mod coupon_mapping;
+pub mod cubic_interpolation;
+pub mod estimator;
+pub mod harmonic_numbers;
+pub mod hash_set;
+pub mod list;
+pub mod serialization;
+pub mod sketch;
```
Xuanwo: Do we need to publish all of them? Some of them feel more like internal implementation details.

src/hll/sketch.rs
```diff
@@ -0,0 +1,422 @@
...
+        match &self.mode {
+            Mode::List { list, .. } => list.container().lower_bound(num_std_dev),
+            Mode::Set { set, .. } => set.container().lower_bound(num_std_dev),
+            Mode::Array4(arr) => arr.lower_bound(num_std_dev),
+            Mode::Array6(arr) => arr.lower_bound(num_std_dev),
+            Mode::Array8(arr) => arr.lower_bound(num_std_dev),
+        }
+    }
+
+    pub fn deserialize(bytes: &[u8]) -> io::Result<HllSketch> {
```
tisonkun: For all `serialize`/`deserialize` methods here, since we don't do IO anyway, I suggest we define DataSketches serde error.

We may further optionally integrate with the [`serde`](https://serde.rs/impl-serialize.html) crate but ser/de sketches should work without `serde` dep. (Its usage requires a serializer/deserializer that involves other concepts beyond sketches from/to bytes)

src/hll/sketch.rs
```diff
@@ -0,0 +1,422 @@
...
+
+        Self {
+            lg_config_k,
+            mode: Mode::List { list, hll_type },
+        }
+    }
+
+    pub fn lg_config_k(&self) -> u8 {
+        self.lg_config_k
+    }
```
tisonkun:
```suggestion
    /// Return the configured lg_config_k.
    pub fn lg_config_k(&self) -> u8 {
        self.lg_config_k
    }
```

nit: we can later (or even now) turn on the `#![deny(missing_docs)]` to ensure all public APIs are well documented.

src/hll/sketch.rs
```diff
@@ -0,0 +1,422 @@
...
+    List = 0,
+    Set = 1,
+    Hll = 2,
+}
+
+#[derive(Debug, Clone)]
+pub struct HllSketch {
+    lg_config_k: u8,
+    mode: Mode,
+}
```
tisonkun: So ditto add some docs and examples here perhaps like https://apache.github.io/datasketches-java/8.0.0/org/apache/datasketches/hll/HllSketch.html

tisonkun:
```suggestion
/// A HyperLogLog sketch.
///
/// See the [module level documentation](self) for more.
#[derive(Debug, Clone)]
pub struct HllSketch {
    lg_config_k: u8,
    mode: Mode,
}
```

Since we have the description at module level, we can simply do this.

src/hll/sketch.rs
```diff
@@ -0,0 +1,422 @@
...
+    pub fn lg_config_k(&self) -> u8 {
+        self.lg_config_k
+    }
+
+    /// Update the sketch with a value
+    ///
+    /// This accepts any type that implements `Hash`. The value is hashed
+    /// and converted to a coupon, which is then inserted into the sketch.
+    /// The sketch will automatically promote from List → Set → HLL as needed.
+    pub fn update<T: Hash>(&mut self, value: T) {
```
tisonkun: "List → Set → HLL" looks like internal impl details. Shall we expose them to end users? I'm afraid that users don't know what this sentence means and they should not.

tisonkun:
```suggestion
    pub fn update<T: Hash>(&mut self, value: T) {
```

tisonkun:
Once we are actively reviewing here, I'd prefer we push new local commits rather than local squash. GitHub's "Squash and Merge" should do the magic while keep new commits as is help reviewer to know where to start from the last review.

src/hll/sketch.rs
```diff
@@ -0,0 +1,347 @@
...
+    pub fn serialize(&self) -> SerdeResult<Vec<u8>> {
+        match &self.mode {
+            Mode::List { list, hll_type } => list.serialize(self.lg_config_k, *hll_type),
+            Mode::Set { set, hll_type } => set.serialize(self.lg_config_k, *hll_type),
+            Mode::Array4(arr) => arr.serialize(self.lg_config_k),
+            Mode::Array6(arr) => arr.serialize(self.lg_config_k),
+            Mode::Array8(arr) => arr.serialize(self.lg_config_k),
+        }
+    }
+}
```
tisonkun: If we are dumping the HllSketch to an always growable `Vec<u8>`, I think we are infallible here:

```diff
diff --git a/src/hll/array4.rs b/src/hll/array4.rs
index 0ab2fce..dbcda53 100644
--- a/src/hll/array4.rs
+++ b/src/hll/array4.rs
@@ -21,6 +21,7 @@
 //! When values exceed 4 bits after cur_min offset, they're stored in an auxiliary hash map.
 
 use super::aux_map::AuxMap;
+use crate::error::SerdeResult;
 use crate::hll::estimator::HipEstimator;
 use crate::hll::{get_slot, get_value};
 
@@ -249,7 +250,7 @@ impl Array4 {
         lg_config_k: u8,
         compact: bool,
         ooo: bool,
-    ) -> crate::error::SerdeResult<Self> {
+    ) -> SerdeResult<Self> {
         use crate::error::SerdeError;
         use crate::hll::serialization::*;
         use crate::hll::{get_slot, get_value};
@@ -325,7 +326,7 @@ impl Array4 {
     /// Serialize Array4 to bytes
     ///
     /// Produces full HLL preamble (40 bytes) followed by packed 4-bit data and optional aux map.
-    pub fn serialize(&self, lg_config_k: u8) -> crate::error::SerdeResult<Vec<u8>> {
+    pub fn serialize(&self, lg_config_k: u8) -> Vec<u8> {
         use crate::hll::pack_coupon;
         use crate::hll::serialization::*;
 
@@ -384,7 +385,7 @@ impl Array4 {
             write_u32_le(&mut bytes, offset, coupon);
         }
 
-        Ok(bytes)
+        bytes
     }
 }
 
diff --git a/src/hll/array6.rs b/src/hll/array6.rs
index 27a5997..4b5d8a4 100644
--- a/src/hll/array6.rs
+++ b/src/hll/array6.rs
@@ -21,6 +21,7 @@
 //! This is sufficient for most HLL use cases without needing exception handling or
 //! cur_min optimization like Array4.
 
+use crate::error::SerdeResult;
 use crate::hll::estimator::HipEstimator;
 use crate::hll::{get_slot, get_value};
 
@@ -149,7 +150,7 @@ impl Array6 {
         lg_config_k: u8,
         compact: bool,
         ooo: bool,
-    ) -> crate::error::SerdeResult<Self> {
+    ) -> SerdeResult<Self> {
         use crate::error::SerdeError;
         use crate::hll::serialization::*;
 
@@ -201,7 +202,7 @@ impl Array6 {
     /// Serialize Array6 to bytes
     ///
     /// Produces full HLL preamble (40 bytes) followed by packed 6-bit data.
-    pub fn serialize(&self, lg_config_k: u8) -> crate::error::SerdeResult<Vec<u8>> {
+    pub fn serialize(&self, lg_config_k: u8) -> Vec<u8> {
         use crate::hll::serialization::*;
 
         let k = 1 << lg_config_k;
@@ -243,7 +244,7 @@ impl Array6 {
         // Write packed byte array
         bytes[HLL_BYTE_ARR_START..].copy_from_slice(&self.bytes);
 
-        Ok(bytes)
+        bytes
     }
 }
 
diff --git a/src/hll/array8.rs b/src/hll/array8.rs
index 6d187d5..4912dc8 100644
--- a/src/hll/array8.rs
+++ b/src/hll/array8.rs
@@ -20,6 +20,7 @@
 //! Array8 is the simplest HLL array implementation, storing one byte per slot.
 //! This provides the maximum value range (0-255) with no bit-packing complexity.
 
+use crate::error::SerdeResult;
 use crate::hll::estimator::HipEstimator;
 use crate::hll::{get_slot, get_value};
 
@@ -119,7 +120,7 @@ impl Array8 {
         lg_config_k: u8,
         compact: bool,
         ooo: bool,
-    ) -> crate::error::SerdeResult<Self> {
+    ) -> SerdeResult<Self> {
         use crate::error::SerdeError;
         use crate::hll::serialization::*;
 
@@ -170,7 +171,7 @@ impl Array8 {
     /// Serialize Array8 to bytes
     ///
     /// Produces full HLL preamble (40 bytes) followed by k bytes of data.
-    pub fn serialize(&self, lg_config_k: u8) -> crate::error::SerdeResult<Vec<u8>> {
+    pub fn serialize(&self, lg_config_k: u8) -> Vec<u8> {
         use crate::hll::serialization::*;
 
         let k = 1 << lg_config_k;
@@ -211,7 +212,7 @@ impl Array8 {
         // Write byte array
         bytes[HLL_BYTE_ARR_START..].copy_from_slice(&self.bytes);
 
-        Ok(bytes)
+        bytes
     }
 }
 
diff --git a/src/hll/hash_set.rs b/src/hll/hash_set.rs
index 9859730..22a583f 100644
--- a/src/hll/hash_set.rs
+++ b/src/hll/hash_set.rs
@@ -136,7 +136,7 @@ impl HashSet {
     }
 
     /// Serialize a HashSet to bytes
-    pub fn serialize(&self, lg_config_k: u8, hll_type: HllType) -> SerdeResult<Vec<u8>> {
+    pub fn serialize(&self, lg_config_k: u8, hll_type: HllType) -> Vec<u8> {
         let compact = true; // Always use compact format
         let coupon_count = self.container.len();
         let lg_arr = self.container.lg_size();
@@ -191,6 +191,6 @@ impl HashSet {
             }
         }
 
-        Ok(bytes)
+        bytes
     }
 }
diff --git a/src/hll/list.rs b/src/hll/list.rs
index 5810750..305e38d 100644
--- a/src/hll/list.rs
+++ b/src/hll/list.rs
@@ -98,7 +98,7 @@ impl List {
     }
 
     /// Serialize a List to bytes
-    pub fn serialize(&self, lg_config_k: u8, hll_type: HllType) -> SerdeResult<Vec<u8>> {
+    pub fn serialize(&self, lg_config_k: u8, hll_type: HllType) -> Vec<u8> {
         let compact = true; // Always use compact format
         let empty = self.container.len() == 0;
         let coupon_count = self.container.len();
@@ -149,6 +149,6 @@ impl List {
             }
         }
 
-        Ok(bytes)
+        bytes
     }
 }
diff --git a/src/hll/sketch.rs b/src/hll/sketch.rs
index 51ead14..dfae923 100644
--- a/src/hll/sketch.rs
+++ b/src/hll/sketch.rs
@@ -284,7 +284,7 @@ impl HllSketch {
     }
 
     /// Serializes the HLL sketch to bytes
-    pub fn serialize(&self) -> SerdeResult<Vec<u8>> {
+    pub fn serialize(&self) -> Vec<u8> {
         match &self.mode {
             Mode::List { list, hll_type } => list.serialize(self.lg_config_k, *hll_type),
             Mode::Set { set, hll_type } => set.serialize(self.lg_config_k, *hll_type),
```

notfilippo: Good catch... In the initial implementation I had some code that would [`Write`](https://doc.rust-lang.org/std/io/trait.Write.html) there, so I had the error.

src/hll/estimator.rs
```diff
@@ -0,0 +1,600 @@
...
+/// within the bounds.
+#[repr(u8)]
+pub enum NumStdDev {
+    /// One standard deviation (\~68% confidence interval)
+    StdDev1 = 1,
+    /// Two standard deviations (\~95% confidence interval)
+    StdDev2 = 2,
+    /// Three standard deviations (\~99.7% confidence interval)
+    StdDev3 = 3,
+}
```
tisonkun:
```suggestion
pub enum NumStdDev {
    /// One standard deviation (\~68% confidence interval)
    One = 1,
    /// Two standard deviations (\~95% confidence interval)
    Two = 2,
    /// Three standard deviations (\~99.7% confidence interval)
    Three = 3,
}
```

Since we typically use it as `NumStdDev::Variant`, maybe we can use `NumStdDev::One` over `NumStdDev::StdDev1`? This is different from Go, where Go's "enum" is simulated and cannot be qualified with a prefix.

notfilippo: I personally don't like spelling out numbers (like 1 vs "One"), but in this context I think your comment makes it nicer :) So why not!

tisonkun: Yeah .. `NumStdDev::1` would be better .. But you know it unfortunately can't be an identifier :\

src/error.rs
```diff
@@ -0,0 +1,52 @@
...
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+//! Error types for datasketches operations
+
+use std::fmt;
+
+/// Result type for datasketches serialization / deserialization operations
+pub type SerdeResult<T> = std::result::Result<T, SerdeError>;
```
tisonkun: What if we remove this alias and use `Result<T, SerdeError>` everywhere?

As far as I can say, downstream projects would seldom use `SerdeResult` - they care about the Error and highly possibly convert to their own error type, making SerdeResult less useful.

Take `std::io::Error` as an example; there is no `std::io::Result`. And ditto `serde::se/der::Error`.

notfilippo:
> Take the crate name on crates.io

https://crates.io/crates/datasketches. I have it since I started the cxx binding project a year ago. I think we can wait for some more substance to cut a proper 0.2.0 release (since my last release was 0.1.3).

src/hll/array4.rs
```diff
@@ -0,0 +1,472 @@
...
+        self.estimator
+            .update(self.lg_config_k, old_value, new_value);
+
+        let shifted_new = new_value - self.cur_min;
+
+        // Four cases based on old/new exception status
+        match (raw_stored, shifted_new) {
+            // Case 1: Both old and new are exceptions
+            (AUX_TOKEN, shifted) if shifted >= AUX_TOKEN => {
+                self.aux_map.as_mut().unwrap().replace(slot, new_value);
```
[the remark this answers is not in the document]
notfilippo: While I'm at it, I'll check all `unwrap()` in non-test code and potentially rewrite them as `expect()` if they are invariants.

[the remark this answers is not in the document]
tisonkun: Yeah .. And perhaps `expect()` in tests can be `unwrap()` when the message is redundant, like `.expect("round-trip deserialization should succeed")`/`.expect("file should exist and should be readable")`. When we're developing and reading test output, the line number often tell all the necessary context :P

src/hll/hash_set.rs
```diff
@@ -0,0 +1,196 @@
...
+                // Duplicate found, nothing to do
+                break;
+            }
+
+            // Collision: compute stride and probe next position
+            // Stride is always odd to ensure all slots are visited
+            let stride = ((coupon & KEY_MASK_26) >> self.container.lg_size()) | 1;
+            probe = (probe + stride) & mask;
+            if probe == starting_position {
+                panic!("HashSet full; no empty slots");
```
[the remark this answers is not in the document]
notfilippo: Added a comment on top of the `panic!` (now `unreachable!`) macro calls.

src/hll/aux_map.rs
```diff
@@ -105,7 +105,9 @@ impl AuxMap {
         let index = self.find(slot);
         match index {
             FindResult::Found(_) => {
-                panic!("slot {} already exists in aux map", slot);
+                // Invariant: Array4 always check existance before inserting
+                // a new value on the same slot.
+                unreachable!("slot {} already exists in aux map", slot);
```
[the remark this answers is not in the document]
tisonkun: In the other direction, if `AuxMap` is open to be used by other struct later, we can keep the `Panics` section but instead document in the caller side how the SAFETY condition is held.

Or even make this method fallible - but this can go in a bikeshedding overengineering direction since I can't define a proper fallible interface immediately.

freakyzoidberg:
LGTM - Thanks !

Issue #10, opened 2025-12-14
Generate snapshots programmatically


tisonkun:
Yeah. I think it's technically reasonable to have snapshots for tests. The key point here is how these snapshots get generated, can we reproducibly make them? And how we can modify/update them.

In other words, how the snapshots in datasketches-go gets generated originally?

_Originally posted by @tisonkun in https://github.com/apache/datasketches-rust/pull/1#discussion_r2616336904_

---

@freakyzoidberg told me there are some "tests" which generate those files in the go repository (https://github.com/apache/datasketches-go/blob/f7bc4b1db865c2dd1be9134d8a61eeb8bc24b1c6/hll/hll_sketch_serialization_test.go#L29) I assumed it's similar for the other implementations.

_Originally posted by @notfilippo in https://github.com/apache/datasketches-rust/pull/1#discussion_r2616338736_

---

Great! Then at least we can reuse the Go logic to generate Go snapshot. But I can see that it would require extra engineer effort so I won't block this PR by such potential improvement to avoid (mysterious) binaries as much as possible.

For the Java and C++ snapshot, perhaps @leerho and @AlexanderSaydakov can give some inputs here.

_Originally posted by @tisonkun in https://github.com/apache/datasketches-rust/pull/1#discussion_r2616343650_

---

CPP / Java and Go repo do have some test that generate and cross-test the synopsis from the other repos.

It's vey much convention and quite manual - and as Lee hinted in the other thread we didn't really think about how to scale this with more language (very much M*N issue)

you can find the Java HLL x-check [here](https://github.com/apache/datasketches-java/blob/main/src/test/java/org/apache/datasketches/hll/HllSketchCrossLanguageTest.java) and the cpp ones for [ser](https://github.com/apache/datasketches-cpp/blob/master/hll/test/hll_sketch_serialize_for_java.cpp)/[de](https://github.com/apache/datasketches-cpp/blob/master/hll/test/hll_sketch_deserialize_from_java_test.cpp) there

Also worth noting that not all synopsis are guaranteed to have byte for byte equality (they'll behave the same, logically equivalent from a behavior aspect and are fully serializable/deserializable between language - but not all provide guarantee of idempotency generation when looking at raw bytes - tldr not all rng are seeded - they could I suppose though)

_Originally posted by @freakyzoidberg in https://github.com/apache/datasketches-rust/pull/1#discussion_r2616369040_

---

> very much M*N issue

[...]

PR #21, opened 2025-12-15
chore: check generated snapshots

tisonkun:
This refers to #10 

I found it quite tricky, and we may need some collaboration from other languages as well to make the snapshot generation more sane.

This is only for tracking the general test structure, how the snapshot generator should work, and the integration tests use generated snapshots.

So far, we can check in the snapshot files as is; just ensure they are excluded from the final crate release. And since they are in binary format, either generating it on the fly or downloading it from somewhere (e.g., GitHub raw assets, nightlies.apache.org) isn't that technically different.

The main purpose here, from what I can see, is to improve interoperability regression tests between different language implementations by exposing the functionality to create snapshots. Then we may have a dedicated repo to hold these snapshots, tag them with the generated date, and download them before running the test.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -86,6 +86,11 @@ jobs:
         run: |
           rustup toolchain install ${{ matrix.rust-version }}
           rustup default ${{ matrix.rust-version }}
+      - uses: actions/setup-go@v6
+        with:
+          go-version: stable
+          cache-dependency-path: |
+            tools/snapshots/go/go.sum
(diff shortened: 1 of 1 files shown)


freakyzoidberg:
I wonder if we could move the generation part in a docker, the docker file could be shared (or even copied between repo, it would rarely have to be modified) ?

The docker could be used in CI, and when needed, locally too

This would encapsulate the convention, and I think I'd rather copy once a docker - than having to sync a folder of synopsis, or go down the road of submodules.

PR #24, opened 2025-12-16
chore: check InsufficientData before index access

tisonkun:
We may later build a shared serde utils where deserialize use a cursor like #23 so we can check insufficient data on the fly, and advancing the curosr on reading to avoid hardcode offset.

### DIFF
--- a/src/hll/mod.rs
+++ b/src/hll/mod.rs
@@ -69,8 +69,8 @@ mod list;
 mod serialization;
 mod sketch;
 
-pub use estimator::NumStdDev;
-pub use sketch::HllSketch;
+pub use self::estimator::NumStdDev;
+pub use self::sketch::HllSketch;
 
 /// Target HLL type.
 ///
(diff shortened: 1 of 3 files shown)


src/hll/mod.rs
```diff
@@ -69,8 +69,8 @@ mod list;
 mod serialization;
 mod sketch;
 
-pub use estimator::NumStdDev;
-pub use sketch::HllSketch;
+pub use self::estimator::NumStdDev;
+pub use self::sketch::HllSketch;
```
notfilippo: What's the idea behind this? Do we have tools to enforce this style repo-wide?

tisonkun: I found it when a mod name sometimes conflicts with a dependency name. Using `self::sketch::HllSketch` always unambiguously refers to the proper symbol.

> Do we have tools to enforce this style repo-wide?

Not quite easy. One can define a custom clippy rule or work with ast-grep. But this rule depends on whether the (pub use) symbol is within this crate/module.

PR #26, opened 2025-12-17
test(hll): HllUnion suite cleanup

notfilippo:
Cleanup of the tests while keeping similar (better) coverage + an assertion fix I noticed while rewriting tests.


src/hll/sketch.rs
```diff
@@ -58,8 +58,9 @@ impl HllSketch {
     /// If lg_config_k is not in range [4, 21]
     pub fn new(lg_config_k: u8, hll_type: HllType) -> Self {
         assert!(
```
freakyzoidberg: Definitely for another review but we should be careful to not crash/abort the program because of bad input from the library.

That should be up to the caller to decide what to do.
And the user should not have to be forced to disabled abort on assert because of the lib.


We (almost) never use panic/abort/assert call in the various implementation and relies on exception/errors instead.

notfilippo: Yes, although the panic clearly documented in the constructor. I think the tests are ok as is but we should consider returning a `Result` for invalid parameters in a separate PR!

PR #29, opened 2025-12-19
build: generate snapshots on the fly 

tisonkun:
This closes #21

This refers to #10

### DIFF
--- a/tools/generate_serialization_test_data.py
+++ b/tools/generate_serialization_test_data.py
@@ -1,4 +1,5 @@
 #!/usr/bin/env python3
+
 # Licensed to the Apache Software Foundation (ASF) under one
 # or more contributor license agreements.  See the NOTICE file
 # distributed with this work for additional information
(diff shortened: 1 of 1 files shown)


tools/generate_serialization_test_data.py
tisonkun: We may have a separate shared repo for holding this/these script(s) and the generated snapshots.

See https://github.com/apache/datasketches-rust/issues/10#issuecomment-3663796398.

tisonkun:
cc @leerho @notfilippo @freakyzoidberg 

@AlexanderSaydakov It seems build datasketches-cpp on Windows failed. Before I dig into it, I'd like to know if datasketches-cpp supposed to be compiled on Windows and if there is some common failure cases.

It seems both C++/Java don't run CI tasks on Windows, so another approach is remove windows from test matrix for now until we have a proper snapshot auto generated script/mechanism then reenable it.

leerho (approved):

PR #39, opened 2025-12-22
feat: add xxhash64

PsiACE:
I just found that we've implemented murmur3hash ourselves, while Java also implements xxhash.

- Ported xxhash64 (seeded) to match DataSketches Java behavior.

### DIFF
--- a/datasketches/src/hash/mod.rs
+++ b/datasketches/src/hash/mod.rs
@@ -16,5 +16,8 @@
 // under the License.
 
 mod murmurhash;
+mod xxhash;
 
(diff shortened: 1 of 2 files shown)


tisonkun:
twox-hash has some `oneshot` specific impls that we may leverage - https://github.com/shepmaster/twox-hash/blob/bc5bb80b4857707e0372d2386157b1d31e4441d3/src/xxhash3_64.rs#L30-L55

But first we should use our own impl in some concrete sketches to see the requirement and input pattern.

PR #43, opened 2025-12-23
build: set up workspace layout 

### LINKED ISSUE #42: Manage our project in workspace with xtask
> I'm considering introducing a xtask-style dev tool and perhaps moving the main `datasketches` crate to a member of a workspace. (then we can have something like `cargo x lint` to apply all the rules instead of letting contributors know what is configured in the CI workflow)
> 
> The other reason for adopting a workspace layout is that I'm considering bringing a binary target that can generate snapshots by the Rust implementation so that other languages' implementations can use it to test cross-language compatibility from Rust to theirs.

_Originally posted by @tisonkun in https://github.com/apache/datasketches-rust/issues/33#issuecomment-3682337588_
            

cc @notfilippo

tisonkun:
This closes #42.

A follow-up is to add a binary target for generating rust snapshots.

cc @notfilippo @Xuanwo

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -101,7 +101,7 @@ jobs:
         working-directory: examples
         run: |
           set -x
-          EXAMPLES=$( yq '.bin.[].name' Cargo.toml )
+          EXAMPLES=( hll_update )
           for example in $EXAMPLES; do
             cargo run --bin "$example"
           done


tisonkun:
cc @notfilippo @freakyzoidberg, sorry to bother, but I'd take a last recent try to see if this is the correct time zone :P

notfilippo (approved):
Looks ok on my end. It's tooling so as long as it simple to use and it works...

tisonkun:
> Looks ok on my end. It's tooling so as long as it simple to use and it works...

Thanks for your feedback!

@leerho could you help in merging this patch?

leerho (approved):

PR #48, opened 2025-12-30
chore: centralize DEFAULT_UPDATE_SEED

tisonkun:
We share this definition and should not expose it outside the crate without a new consensus.

cc @leerho


freakyzoidberg (approved):

freakyzoidberg:
LGTM - Thanks

PR #49, opened 2025-12-30
refactor: replace byteorder with SketchBytes and SketchSlice 

### LINKED ISSUE #41: Encapsulate codec over Vec<u8> and slices
> On a separate note, re: removing dependencies. What do we feel about `byteorder`?

I think we can have our own reiml perhaps in `mod serde` or find a better name. We can optimize our use cases over `Vec<u8>` and `&[u8]`, while both BE and LE are needed - we still use BigEndian somewhere, like recognizing TDigests' reference format.

Let me track this task and try to implement it this week.

_Originally posted by @tisonkun in https://github.com/apache/datasketches-rust/issues/36#issuecomment-3682345690_

tisonkun:
This closes #41.

I'd wait for a review from @notfilippo @Xuanwo 

cc @PsiACE @ZENOTME

### DIFF
--- a/datasketches/src/codec.rs
+++ b/datasketches/src/codec.rs
@@ -18,7 +18,8 @@
 #![allow(dead_code)]
 
 use std::io;
-use std::io::{Cursor, Read};
+use std::io::Cursor;
+use std::io::Read;
 
 pub(crate) struct SketchBytes {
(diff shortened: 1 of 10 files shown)


ZENOTME (approved):
LGTM

PR #51, opened 2025-12-31
chore: fine tune frequencies and theta sketches

tisonkun:
cc @PsiACE @ZENOTME

### DIFF
--- a/datasketches/src/theta/mod.rs
+++ b/datasketches/src/theta/mod.rs
@@ -32,6 +32,5 @@
 mod hash_table;
 mod sketch;
 
-pub use self::hash_table::ResizeFactor;
 pub use self::sketch::ThetaSketch;
 pub use self::sketch::ThetaSketchBuilder;
--- a/datasketches/src/lib.rs
+++ b/datasketches/src/lib.rs
@@ -39,3 +39,6 @@ pub mod theta;
 
 mod codec;
 mod hash;
+mod resize;
+
+pub use self::resize::ResizeFactor;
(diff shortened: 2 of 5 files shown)


datasketches/src/frequencies/mod.rs
```diff
@@ -29,7 +29,6 @@ mod serde;
 mod serialization;
 mod sketch;
 
-pub use self::serde::ItemsSerde;
```
tisonkun: To @PsiACE,

No need `ItemsSerde` if we already implement serde over concrete `FrequentItemsSketch<i64>` and `FrequentItemsSketch<String>`.

But the C++ impl does have a Serde abstraction you may investigate in https://github.com/apache/datasketches-cpp/blob/master/common/include/serde.hpp

datasketches/src/theta/mod.rs
```diff
@@ -32,4 +32,6 @@
 mod hash_table;
 mod sketch;
 
+pub use self::hash_table::ResizeFactor;
 pub use self::sketch::ThetaSketch;
+pub use self::sketch::ThetaSketchBuilder;
```
tisonkun: To @ZENOTME,

`ThetaSketchBuilder` and `ResizeFactor` are needed because `ThetaSketch::builder` exposes the builder, which has a `resize_factor` method that only accepts a `ResizeFactor` instance.

datasketches/src/frequencies/mod.rs
tisonkun: cc @AlexanderSaydakov I'd appreciate it if you can share the story and usage of `serde.hpp`.

PsiACE (approved):
lgtm 💯 on frequencies sketch.

Xuanwo (approved):
Seems nice!

PR #58, opened 2025-12-31
docs: add rustdoc examples across sketches

### LINKED ISSUE #55: Add rustdoc examples for public API
Currently only some sketches (e.g. [bloom](https://github.com/apache/datasketches-rust/blob/819f563c29cfd139a0a69be052591a5aa3368e4d/datasketches/src/bloom/sketch.rs#L63-L72)) expose rustdoc examples. We should update most public API to have examples to make the generated documentation interactive, easy to digest and LLM-friendly.

We can split this issue into separate tasks for each sketch if necessary but a good pass of the current public API should be doable in a single PR.

PsiACE:
- Add rustdoc examples across countmin/frequencies/hll/theta/tdigest
- Add minimal examples for public Error and ResizeFactor types to improve doc interactivity.

Close #55

### DIFF
--- a/datasketches/src/countmin/sketch.rs
+++ b/datasketches/src/countmin/sketch.rs
@@ -57,7 +57,7 @@ impl CountMinSketch {
     ///
     /// # Examples
     ///
-    /// ```
+    /// ```rust
     /// # use datasketches::countmin::CountMinSketch;
     /// let sketch = CountMinSketch::new(4, 128);
     /// assert_eq!(sketch.num_buckets(), 128);
@@ -75,7 +75,7 @@ impl CountMinSketch {
     ///
     /// # Examples
     ///
-    /// ```
+    /// ```rust
     /// # use datasketches::countmin::CountMinSketch;
     /// let sketch = CountMinSketch::with_seed(4, 64, 42);
     /// assert_eq!(sketch.seed(), 42);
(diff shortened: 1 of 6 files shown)


leerho:
I need to see an approving review by a Rust contributor before I feel comfortable in merging this  :)

datasketches/src/countmin/sketch.rs
```diff
@@ -195,6 +252,16 @@ impl CountMinSketch {
...
     /// Serializes this sketch into the DataSketches Count-Min format.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// # use datasketches::countmin::CountMinSketch;
+    /// # let mut sketch = CountMinSketch::new(4, 128);
+    /// # sketch.update("apple");
+    /// let bytes = sketch.serialize();
+    /// let _decoded = CountMinSketch::deserialize(&bytes).expect("valid bytes");
```
notfilippo: As per [Rust common message conventions](https://doc.rust-lang.org/std/error/index.html#common-message-styles), expect should describe verbosely what we would expect from the operation (usually with a "should" word somewhere in the phrase). 

I would argue that for the sake of the examples we should directly use unwrap.

Ditto for every instance of `expect`.

notfilippo (approved):
Looks good. Just a couple of nits.

leerho (approved):

PR #60, opened 2026-01-01
feat: try to intro a Xor8 impl

PsiACE:
- This is just a simple attempt to implement xor8. For evaluation purposes.


tisonkun:
ref -

https://arxiv.org/pdf/1912.08258
https://github.com/prataprc/xorfilter
https://github.com/FastFilter/xorfilter

PR #57, opened 2025-12-31
chore: prepare v0.2.0 release

notfilippo:
Closes #56 .

PR #61, opened 2026-01-02
docs: document release process

notfilippo:
Related to #57 .

Attempt to document a potential release process.

cc @leerho @freakyzoidberg @tisonkun

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,301 @@
+<!--
+    Licensed to the Apache Software Foundation (ASF) under one
+    or more contributor license agreements.  See the NOTICE file
+    distributed with this work for additional information
+    regarding copyright ownership.  The ASF licenses this file
+    to you under the Apache License, Version 2.0 (the
+    "License"); you may not use this file except in compliance
+    with the License.  You may obtain a copy of the License at
+
+      http://www.apache.org/licenses/LICENSE-2.0
+
+    Unless required by applicable law or agreed to in writing,
+    software distributed under the License is distributed on an
+    "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+    KIND, either express or implied.  See the License for the
+    specific language governing permissions and limitations
+    under the License.
+-->
+
+# Release process for Rust components
+
+**NOTES:**
+
+* This process covers major and minor releases only. Bug-fix releases, which increment the third digit, are performed on
+  an A.B.X branch and not on main, but otherwise follow a similar process.
+* Some of these operations can be performed either on the Command-Line or in your IDE, whatever you prefer.
+
+## Preparation
+
+### Verify project health
+
+* Confirm correctness for:
+    * LICENSE
+    * NOTICE -- check for copyright dates
+    * README.md
+    * .asf.yaml
+    * .gitignore
+    * Cargo.toml -- ensure workspace metadata is correct
+
+* From Command Line or IDE:
+    * Run all checks: `cargo x lint`
+    * Run unit tests: `cargo x test`
+    * Run `cargo update` to ensure Cargo.lock is updated
+    <!-- TODO: * Run code coverage: `cargo llvm-cov --workspace` (target > 90%). -->
+    * Run tests on all platforms (see CI workflow)
+    * Confirm that documentation builds: `cargo doc --open`
+    * Confirm that all **temporary** branches are checked into main and/or deleted, both local and remote
+    * Confirm any new bug fixes have corresponding tests
+    * Check that all dependencies are at stable versions (no pre-release dependencies in Cargo.toml)
+
+### Verify release tooling
+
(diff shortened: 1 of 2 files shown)


RELEASE.md
```diff
@@ -0,0 +1,301 @@
...
+
+* Check and grab the web URL: *https://dist.apache.org/repos/dist/dev/datasketches/rust/A.B.X-rc.1/*
+    * There should be 3 files: \*-src.zip, \*-src.zip.asc, \*-src.zip.sha512
+    * Verify signatures:
+        * `gpg --verify datasketches-rust-A.B.X-rc.1-src.zip.asc datasketches-rust-A.B.X-rc.1-src.zip`
+        * `sha512sum -c datasketches-rust-A.B.X-rc.1-src.zip.sha512`
+
+## Prepare & Send [VOTE] Letter to dev@
+
+* See VoteTemplates directory for a recent example
```
notfilippo: Not sure where `VoteTemplates` is but it is mentioned on both Java and CPP. cc @leerho

RELEASE.md
```diff
@@ -0,0 +1,301 @@
...
+    * GitHub Tag: A.B.X-rc.1 (or rc.N)
+
+* Start a new terminal in the *dist/dev/datasketches/scripts* directory on your system:
+    * To confirm *gpg-agent* is running type:
+        * `ps -axww | grep gpg`  # you should see something like:
+            * *64438 ?? 0:30.33 gpg-agent --homedir /Users/\<name\>/.gnupg --use-standard-socket --daemon*
+        * To start GPG if GPG Agent is not running:
+            * `eval $(gpg-agent --daemon)`
+    * Run the deployment script:
+        * `./bashDeployToDist.sh /Users/\<name\>/dev/git/Apache/datasketches-rust datasketches-rust A.B.X-rc.1`
```
tisonkun: > bashDeployToDist.sh

Do we have this script now?

tisonkun:
I'm feeling this is too complicated to follow. Although it covers many details, may we start with an easier version?

That is, many works should be guarded with CI. The release only steps are packaging the source tarball + calling a release on the mailing list. Perhaps with an extra cargo release, which can be automated later, like https://github.com/apache/opendal/blob/b430879d45412e960f2bb8313558dbc18feec6c9/.github/workflows/release_rust.yml#L76-L82

cc @Xuanwo

RELEASE.md
```diff
@@ -0,0 +1,301 @@
...
+
+## Create permanent release branch & release candidate version preparation
+
+* From IDE or Command Line:
+    * Create new **Permanent Branch**: "A.B.X"
+    * Modify `datasketches/Cargo.toml` version:
+        * Change version to `version = "A.B.X"`
+    * Commit the changes with message: "chore: prepare A.B.X release"
+    * Write down the Git hash: example: 40c6f4f
+    * Push Branch "A.B.X" to origin:
```
tisonkun: I don't think we will cut a new branch for each release now. But simply has tags and a linear history on the main branch.

We should use `A.B.X-rc.N` during release candidates; here it writes "A.B.X" without the rc suffix.

RELEASE.md
```diff
@@ -0,0 +1,301 @@
...
+    * .gitignore
+    * Cargo.toml -- ensure workspace metadata is correct
+
+* From Command Line or IDE:
+    * Run all checks: `cargo x lint`
+    * Run unit tests: `cargo x test`
+    * Run `cargo update` to ensure Cargo.lock is updated
+    <!-- TODO: * Run code coverage: `cargo llvm-cov --workspace` (target > 90%). -->
+    * Run tests on all platforms (see CI workflow)
+    * Confirm that documentation builds: `cargo doc --open`
```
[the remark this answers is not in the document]
notfilippo: Can we have an X script for this?

RELEASE.md
PsiACE: At Opendal, we've switched some mailing list processes to GitHub Discussions as the primary channel and established a standardized format. Like this https://github.com/apache/opendal/discussions/6794

notfilippo: That's interesting. I'm in favour of this but I'm new to the ASF way of doing things so maybe @leerho can have a more informed opinion.

tisonkun (approved):
I think we're generally fine here. We can review and polish the document during our first ASF release. Typically, RELEASE.md is a materialized copy of what we really do then.

If it's desired, you can add me as a crate owner to the `datasketches` crate and I can participate in the release process.

Generally, we need a PMC member as the acting release manager and at least three binding votes from DataSketches PMC members to call a release.

Other improvements, like using GH Discussions as the major channel while keeping the list informed, and automating the release process, can be follow-ups.

notfilippo:
I've sent an invite @tisonkun !

I will release an rc.1 shortly!

leerho (approved):
This is a live document as it will continue to have revisions.

PR #66, opened 2026-01-08
chore: Remove required approving review count from PRs

tisonkun:
Add it back when we have more committers to extend the review bandwidth.

cc @leerho @freakyzoidberg 

This is a proposal. I hope to cut the new release candidate this week and I'd ask for relaxing this setting for about one month during nominating new committers.

### DIFF
--- a/.asf.yaml
+++ b/.asf.yaml
@@ -38,8 +38,7 @@ github:
       required_status_checks:
         contexts:
           - Required
-      required_pull_request_reviews:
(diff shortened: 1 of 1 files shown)


jmalkin:
I would suggest leaving policy changes open for discussion for a longer period of time than this was.

Given the increasing number of attacks on open source software, I tend to pretty strongly support the extra check even though it definitely slows things down at times. Just as a way of giving users of the library additional confidence in our process.

Issue #70, opened 2026-01-14
Support halve and decay for CountMinSketch


tisonkun:
Ref -

* https://docs.rs/cmsketch/latest/cmsketch/struct.CMSketchU16.html

While migrating @mrcroxx's foyer from CMSketch to DataSketches' CountMinSketch, I noticed this gap.

Our implementation was initialized at:

* https://github.com/apache/datasketches-rust/pull/38
* https://github.com/apache/datasketches-java/issues/494
* https://github.com/apache/datasketches-go/pull/42
* https://github.com/apache/datasketches-cpp/pull/325

I wonder if `halve` and `decay` are part of the standard CountMin sketch and how we can implement it into datasketches-rust.

cc @PsiACE @jmalkin @c-dickens @geonove as original authors

tisonkun:
Resolved by #71.

PR #71, opened 2026-01-14
feat: add halve and decay to countmin sketch of unsigned values

PsiACE:
Added `halve` and `decay` to CountMinSketch, matching cmsketch behavior for non‑negative counters. 

This scales counters (and `total_weight`) to support exponential decay while keeping serialization unchanged. I’m not fully confident about the math details, especially around our error bounds, but the implementation itself should be straightforward. 😃 

Related #70

### DIFF
--- a/datasketches/src/countmin/sketch.rs
+++ b/datasketches/src/countmin/sketch.rs
@@ -172,12 +172,12 @@ impl<T: CountMinValue> CountMinSketch<T> {
             return;
         }
         let abs_weight = weight.abs();
-        self.total_weight = self.total_weight.wrapping_add(abs_weight);
+        self.total_weight = self.total_weight.add(abs_weight);
         let num_buckets = self.num_buckets as usize;
         for (row, seed) in self.hash_seeds.iter().enumerate() {
             let bucket = self.bucket_index(&item, *seed);
             let index = row * num_buckets + bucket;
-            self.counts[index] = self.counts[index].wrapping_add(weight);
+            self.counts[index] = self.counts[index].add(weight);
         }
     }
 
@@ -214,7 +214,7 @@ impl<T: CountMinValue> CountMinSketch<T> {
     pub fn upper_bound<I: Hash>(&self, item: I) -> T {
         let estimate = self.estimate(item);
         let error = T::from_f64(self.relative_error() * self.total_weight.to_f64());
-        estimate.wrapping_add(error)
+        estimate.add(error)
     }
 
     /// Merges another sketch into this one.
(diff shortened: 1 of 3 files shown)


c-dickens:
To answer the question directly: this example of scaling up or down by a **nonnegative** value is acceptable.  However, there is a knock-on effect on the error bound which needs testing.  Basically, for an underlying frequency value `f_i` of item `i` the error bound is `true(f_i) <= est(f_i) <= true(f_i) + error` where error depends on the sketch size and the total mass in the stream.  So if you are scale the inputs to the sketch (that make up each "true" `f_i`), then you also need to scale the `total_weight` - this is implemented in the PR but the altered error bounds are not tested. 

In future, we do need a little care because there is some subtlety around the behaviour depending on the input stream.  If only **positive** integers are used for addition and multiplication **and** the underlying stream only has positive frequencies, then things are ok.  The nuance comes in around negative values in the underlying frequency vector, in which case we might prefer a `CountSketch` -- something useful that we haven't got round to yet!

There are some good notes [here](https://fuhuthu.com/BigData2022/countmin.pdf) and [here](https://www.cs.rice.edu/~as143/COMP480_580_Spring20/scribe/lect10.pdf) that make the point clear.  The "standard" count min point query estimation is only valid on the assumption that all entries in the underlying frequency vector are nonnegative.  Happy to give more details if need be.

In reference to @tisonkun's comment, there are a bunch of windowing/merging/aggregate approaches that might be interested.  If so please add to the [discussion](https://github.com/apache/datasketches-rust/discussions/64) - I am very keen to learn if you have specific applications in mind, or whether they (and other sketches eg. CountSketch) would be generally useful.

PsiACE:
Marking this as ready for review. I added a simple test to make sure the scaling behavior basically works (single key, no collisions). We can use that as a starting point

c-dickens:
@leerho - I have enquired with Justin to check my understanding.  For others, my feeling is that halving (or scaling by some factor in (0, 1] ) is an acceptable operation.  However, the sketch array has signed values, so if sketches were subtracted from one another and a negative value became present, then this would violate the error bounds.  The standard (and well-known) error bounds for count min frequency estimation have a key assumption that all entries in the underlying frequency vector are non-negative so a negative value in `sketch(A) - sketch(B)` would contradict this assumption and thus the error bounds would not be valid.  

Although the scaling itself is not a problem, it would be good to learn more about the intended application to decide the best path forward.

PsiACE:
> I'm considering to generify `CountMinSketch` to accept `T: IntNumber` and only implement halve/decay for unsigned versions.
> 
> This should be the same as [cmsketch](https://docs.rs/cmsketch), which implements only for unsigned values.

I agree. I'll get back to the code later.

PR #86, opened 2026-02-08
refactor: sort PairTable values with std sort_unstable

tisonkun:
The std sort_unstable algorithm already leverages existing sorted elements.

cc @ZENOTME @notfilippo @PsiACE
fyi @leerho


PsiACE (approved):
Nice. lgtm

Issue #89, opened 2026-02-11
Inconsistent BloomFilter.MAX_SIZE_BITS


tisonkun:
In [datasketches-java](https://github.com/apache/datasketches-java/blob/3def49b3ea13c779a3aef81f9ca8c54d8013a62d/src/main/java/org/apache/datasketches/filters/bloomfilter/BloomFilter.java#L65):

```java
public static final long MAX_SIZE_BITS = (Integer.MAX_VALUE - Family.BLOOMFILTER.getMaxPreLongs()) * (long) Long.SIZE;
// result in (Integer.MAX_VALUE - 4) * 64 = 137_438_953_152
```

In [datasketches-cpp](https://github.com/apache/datasketches-cpp/blob/fccb2385f3a66416fc34c0d7bd0513696721ecd9/filters/include/bloom_filter.hpp#L577-L581):

```cpp
static const uint64_t MAX_HEADER_SIZE_BYTES = 32; // 4 Java Longs
static const uint64_t MAX_FILTER_SIZE_BITS = (INT32_MAX - MAX_HEADER_SIZE_BYTES) * sizeof(uint64_t);
// result in (INT32_MAX - 32) * 8 = 17_179_868_920
```

In [datasketches-go](https://github.com/apache/datasketches-go/blob/2cb3a8c970d797cf95d1efc4aa28b7cff38a29a8/filters/bloom_filter_builder.go#L62-L65):

```go
maxBits := uint64(math.MaxInt32-32) * 64
// result in 137_438_951_360
```

In [datasketches-rust](https://github.com/apache/datasketches-rust/blob/9a830707410a66823e997b04cfd6e48890904957/datasketches/src/bloom/builder.rs#L22):

```rust
const MAX_NUM_BITS: u64 = (1u64 << 35) - 64; // ~32 GB - reasonable limit 
// result in 34_359_738_304
```

None of them is equal.

Besides, in datasketches-rust, there is a lower bound (`MIN_NUM_BITS`) = 64, while other impls check only `num_bits > 0`.

@PsiACE where is the bound of bloomfilter comes from in datasketches-rust?

@leerho @AlexanderSaydakov @freakyzoidberg @proost You may take a look and I wonder which one of the upper bound is the correct one.

jmalkin:
Java is correct. Max array length in Java is 2^32-1, the longest contiguous array is of primitive longs, and there’s a little room taken for the preamble in the actual serialized sketch.

[...]

PR #91, opened 2026-02-13
refactor: expose codec and add centralized Family 

### LINKED ISSUE #67: Centralize several serde constants
... perhaps to the `datasketches::codec` module, or a new internal `consts`/`serde` module.

This follows what @proost does in https://github.com/apache/datasketches-go/pull/97 that I think a good way to go.

So far, we store these constants in `datasketches::<sketch>::serde` path but we'd better manage them (at least the family IDs) in a centralized place.

tisonkun:
This closes #67.

Later I'll

* Expose `FrequentItemValue` trait since `codec::SketchSlice` and `codec::SketchBytes` are public now.
* Use `max_pre_longs` to calculate #89.

### DIFF
--- a/datasketches/src/bloom/sketch.rs
+++ b/datasketches/src/bloom/sketch.rs
@@ -449,11 +449,6 @@ impl BloomFilter {
         }
 
         let is_empty = (flags & EMPTY_FLAG_MASK) != 0;
-        if is_empty {
-            debug_assert_eq!(preamble_longs, Family::BLOOMFILTER.min_pre_longs);
-        } else {
-            debug_assert_eq!(preamble_longs, Family::BLOOMFILTER.max_pre_longs);
-        }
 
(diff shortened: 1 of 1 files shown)


datasketches/src/bloom/sketch.rs
```diff
@@ -432,27 +430,30 @@ impl BloomFilter {
...
+                "invalid preamble longs: expected [{}, {}], got {}",
+                Family::BLOOMFILTER.min_pre_longs,
+                Family::BLOOMFILTER.max_pre_longs,
+                preamble_longs
+            )));
         }
 
         let is_empty = (flags & EMPTY_FLAG_MASK) != 0;
+        if is_empty {
+            debug_assert_eq!(preamble_longs, Family::BLOOMFILTER.min_pre_longs);
```
notfilippo: Why do we use `debug_assert_eq` here? Are we worried about performance?

PR #92, opened 2026-02-13
test: add count-min C++ serialization compatibility tests

PsiACE:


tisonkun:
Seems there are no snapshots of other impls. Perhaps you can start a new thread to coordinate a collaboration between different impls.

tisonkun:
Failed on Windows. Perhaps related to line separator?

tisonkun:
Looks like when generating the cpp snapshot, it prints CRLF on windows.

Should work with their people like @proost

leerho:
WRT Windows and CRLF issues.  I noticed that there is no .gitattributes file in ds-cpp.  I ran into similar issues recently when refactoring ds-memory.  It turns out all the major modern OSs can correctly process text files with just LF "\n" line endings.  So structuring the .gitattributes file to force LF endings for all text type files fixed the problem and it works for Ubuntu, MacOS, and Windows.  See this .[gitattributes](https://github.com/apache/datasketches-memory/blob/main/.gitattributes).

tisonkun:
@PsiACE the corresponding cpp patch has been merged.

You may rebase this patch and pin the cpp checkout to the latest commit and retry.

PsiACE:
Rebased this branch onto current `origin/main` and pinned the C++ serialization data generator to apache/datasketches-cpp@0bab2596260510a40c1c953af1d6746c19d87dd1, which includes the merged .gitattributes update.

tisonkun:
Seems C++ snapshot generation issue, not gitattribute.

Let me prepare a fix.

PsiACE:
> Seems C++ snapshot generation issue, not gitattribute.
> 
> Let me prepare a fix.

`*.bin` -> `*.sk` , and with  `std::ios::binary` , emmmm, it looks like this should fix the problem.

tisonkun:
See https://github.com/apache/datasketches-cpp/pull/505.

tisonkun:
You may test with checkout to this patch's commit so we use it as an e2e test.

tools/generate_serialization_test_data.py
```diff
@@ -166,13 +168,22 @@ def generate_cpp_files(workspace_dir, project_root):
...
         files_copied += 1
 
+    # Count-Min test binaries are produced as `count_min-*.bin`.
+    # Normalize names to match the repository snapshot convention.
+    for file_path in build_dir.rglob("count_min-*.bin"):
+        base_name = file_path.stem[len("count_min-"):].replace("-", "_")
+        output_name = f"countmin_{base_name}_cpp.sk"
+        shutil.copy2(file_path, output_dir / output_name)
+        print(f"Copied: {output_name} (from {file_path.name})")
+        files_copied += 1
```
tisonkun: I don't think we need this hack. We should use `count_min-*.bin` in the Rust test code.

It doesn't matter that we just follow what cpp uses. If we want to unify the name conventions, we should unify from the source side, not arbitrarily patching within downstream.

PsiACE: good catch. and fixed with `for pattern in ("*_cpp.sk", "count_min-*.bin"):` , no rename.

tools/generate_serialization_test_data.py
```diff
@@ -142,6 +146,8 @@ def generate_cpp_files(workspace_dir, project_root):
         repo_url,
         str(temp_dir)
     ])
+    run_command(["git", "fetch", "--depth", "1", "origin", fetch_ref], cwd=temp_dir)
+    run_command(["git", "checkout", "--detach", commit], cwd=temp_dir)
```
PsiACE, line 150: some temp code here; needs to be cleaned once PR505 merged into c++ impl

tisonkun (approved):
LGTM - merging now.

We can drop python scripts trick once datasketches-cpp updated.

PR #93, opened 2026-02-13
fix(bloom): align builder parameters with java implementation

### LINKED ISSUE #89: Inconsistent BloomFilter.MAX_SIZE_BITS

notfilippo:
Closes #89.

### DIFF
--- a/datasketches/src/bloom/builder.rs
+++ b/datasketches/src/bloom/builder.rs
@@ -19,11 +19,6 @@ use super::BloomFilter;
 use crate::codec::family::Family;
 use crate::hash::DEFAULT_UPDATE_SEED;
 
-pub const MIN_NUM_BITS: u64 = 1;
-pub const MAX_NUM_BITS: u64 = (i32::MAX as u64 - Family::BLOOMFILTER.max_pre_longs as u64) * 64;
-pub const MIN_NUM_HASHES: u16 = 1;
-pub const MAX_NUM_HASHES: u16 = i16::MAX as u16;
-
 /// Builder for creating [`BloomFilter`] instances.
 ///
 /// Provides two construction modes:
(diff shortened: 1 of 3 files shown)


tisonkun (changes requested):
Generally looks good. Two comments here:

1. The `PREAMBLE_LONGS_STANDARD` can migrate to `Family::BLOOMFILTER.max_pre_longs` after #91 merged.
2. In the Java impl there is no `MIN_NUM_BITS = 64`. It checks only `num_bits > 0` (strictly positive). (Logically `MIN_NUM_BITS = 1` I think)

datasketches/src/bloom/builder.rs
```diff
@@ -88,8 +94,8 @@ impl BloomFilterBuilder {
     /// # Panics
     ///
     /// Panics if any of:
-    /// - `num_bits` < MIN_NUM_BITS (64) or `num_bits` > MAX_NUM_BITS (~32 GB)
-    /// - `num_hashes` < 1 or `num_hashes` > 100
+    /// - `num_bits` < MIN_NUM_BITS or `num_bits` > MAX_NUM_BITS
+    /// - `num_hashes` < MIN_NUM_HASHES or `num_hashes` > MIN_NUM_HASHES
```
tisonkun: Constants are not included in the final public documents. We'd better to speak out the exact value here.

tisonkun (commented):
LGTM. A few nice-to-handle comments.

tisonkun (approved):

datasketches/src/bloom/builder.rs
tisonkun: Reasonable.

tisonkun: This should be OK.

PR #95, opened 2026-02-14
chore(x): add doc lint

notfilippo:
Add `cargo doc` that treats warnings as errors to the linting suite of the xtast.


tisonkun (approved):
Generally looks good.

But since it cannot be fixed anyway, I'd prefer run the same warning as error when executing linters.

This is because during the development, developers may often run only `cargo x lint --fix`, so error in the first place helps.

PR #100, opened 2026-02-19
feat(theta): introduce intersection theta set operation

ZENOTME:
This PR:
- introduce ThetaSketchView trait and enhance ThetaHashTable used for set operation
- introduce intersection theta set operation


datasketches/src/theta/hash_table.rs
```diff
@@ -302,6 +343,55 @@ impl ThetaHashTable {
         compute_seed_hash(self.hash_seed)
     }
 
+    /// Returns true if the given hash exists in the table.
+    #[allow(dead_code)]
+    pub fn contains_hash(&self, hash: u64) -> bool {
```
tisonkun: Why allow dead_code?

I'd prefer to add private methods with their usage instead of pre-inject logics.

ZENOTME: done

tisonkun (commented):
I suggest you directly implement all the requirements (set ops).

These abstractions look intermediate - I can't judge whether they are correct isolated.

datasketches/src/theta/hash_table.rs
```diff
@@ -77,34 +83,58 @@ impl ThetaHashTable {
...
+            true,
+        )
+    }
+
+    /// Constructs a table from raw internal state.
+    ///
+    /// # Panics
+    ///
+    /// Panics if `lg_cur_size > lg_nom_size + 1`. (`lg_nom_size + 1 == lg_max_size`)
+    pub fn from_raw_parts(
```
ZENOTME, line 92: This function is mainly used in intersection, and the current implementation is the same as C++. Use the builder pattern, maybe clearer. E.g.
```
// 
let hash_table = 
      ThetaHashTable::builder(lg_nom_size, resize_factor, hash_seed)
            .with_sampling_probability(...) // infer the theta from sampling_probability
            .with_log_cur_size(...); // set log_cur_size manually, ensure log_cur_size is valid. If not set, the log cur size will be inferred from log_nom_size.
```
cc @tisonkun

ZENOTME:
> I suggest you directly implement all the requirements (set ops).
> 
> These abstractions look intermediate - I can't judge whether they are correct isolated.

Have push commit for intersection op

tisonkun (approved):
Thanks for your contribution @ZENOTME! LGTM.

Add commits to seal the `ThetaSketchView` trait.

PR #111, opened 2026-03-24
fix: add SketchHashable trait to fix inconsistent hashing across types

ZENOTME:
`XXXSketch::update` previously accepted arbitrary `T: Hash` and fed Rust-native Hash semantics directly into the hash path. This diverges from the Java/C++ datasketches implementations, which use explicit canonical update semantics for numeric, floating-point, string, and byte-oriented inputs.

Concretely, this caused several cross-language mismatches:
- narrow integers were not canonicalized through the same signed-widening path as Java/C++
- str / byte-oriented inputs followed Rust structural Hash semantics instead of raw-byte semantics
- the default integer-stream path (`for example for i in 0..n { sketch.update(i) }`) could produce different retained hashes / estimates from the C++ implementation

This PR introduce SketchHashable trait to unify the hash behaviour to fix inconsistent hashing across types

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -168,13 +168,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "examples"
-version = "0.0.0"
-dependencies = [
- "datasketches",
-]
-
-[[package]]
 name = "fastrand"
 version = "2.3.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
--- a/datasketches/src/cpc/sketch.rs
+++ b/datasketches/src/cpc/sketch.rs
@@ -24,7 +24,6 @@ use crate::codec::assert::ensure_serial_version_is;
 use crate::codec::assert::insufficient_data;
 use crate::codec::family::Family;
 use crate::common::NumStdDev;
-use crate::common::canonical_double;
 use crate::common::inv_pow2_table::INVERSE_POWERS_OF_2;
 use crate::cpc::DEFAULT_LG_K;
 use crate::cpc::Flavor;
@@ -49,6 +48,7 @@ use crate::error::Error;
 use crate::error::ErrorKind;
 use crate::hash::DEFAULT_UPDATE_SEED;
 use crate::hash::MurmurHash3X64128;
+use crate::hash::SketchHashable;
 use crate::hash::compute_seed_hash;
 
 /// A Compressed Probabilistic Counting sketch.
@@ -170,12 +170,10 @@ impl CpcSketch {
         self.num_coupons == 0
     }
 
-    /// Update the sketch with a hashable value.
-    ///
-    /// For `f32`/`f64` values, use `update_f32`/`update_f64` instead.
-    pub fn update<T: Hash>(&mut self, value: T) {
+    /// Update the sketch with a value that implements [`SketchHashable`].
+    pub fn update<T: SketchHashable>(&mut self, value: T) {
         let mut hasher = MurmurHash3X64128::with_seed(self.seed);
-        value.hash(&mut hasher);
+        value.to_hashable().hash(&mut hasher);
         let (h1, h2) = hasher.finish128();
 
         let k = 1 << self.lg_k;
@@ -189,18 +187,6 @@ impl CpcSketch {
             row_col ^= 1 << 6;
         }
         self.row_col_update(row_col);
-    }
-
-    /// Update the sketch with a f64 value.
-    pub fn update_f64(&mut self, value: f64) {
-        // Canonicalize double for compatibility with Java
-        let canonical = canonical_double(value);
-        self.update(canonical);
-    }
-
-    /// Update the sketch with a f32 value.
-    pub fn update_f32(&mut self, value: f32) {
-        self.update_f64(value as f64);
     }
 
     pub(super) fn flavor(&self) -> Flavor {
--- a/datasketches/src/cpc/union.rs
+++ b/datasketches/src/cpc/union.rs
@@ -125,11 +125,11 @@ impl CpcUnion {
     /// # use datasketches::cpc::CpcSketch;
     ///
     /// let mut s1 = CpcSketch::new(12);
-    /// s1.update(&"apple");
+    /// s1.update("apple");
     ///
     /// let mut s2 = CpcSketch::new(12);
-    /// s2.update(&"apple");
-    /// s2.update(&"banana");
+    /// s2.update("apple");
+    /// s2.update("banana");
     ///
     /// let mut union = CpcUnion::new(12);
     /// union.update(&s1);
--- a/datasketches/src/hash/mod.rs
+++ b/datasketches/src/hash/mod.rs
@@ -15,10 +15,14 @@
 // specific language governing permissions and limitations
 // under the License.
 
+//! Shared hashing utilities.
+
 mod murmurhash;
+mod sketch_hashable;
 mod xxhash;
 
 pub(crate) use self::murmurhash::MurmurHash3X64128;
+pub use self::sketch_hashable::SketchHashable;
 pub(crate) use self::xxhash::XxHash64;
 
 /// The seed 9001 used in the sketch update methods is a prime number that was chosen very early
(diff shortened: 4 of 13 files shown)


ZENOTME:
Seems CPC update and HLL update may have the same problem. cc @tisonkun @PsiACE

tisonkun (changes requested):
I don't get the point clear. Could you provide a failing test to show how the original impl is wrong?

ZENOTME:
> I don't get the point clear. Could you provide a failing test to show how the original impl is wrong?

e.g. for u32 in Rust, the following test fails before this PR
```
fn test_unsigned_narrow_integer_inputs_follow_cpp_signed_path() {
      let mut u32_sketch = ThetaSketch::builder().build();
      u32_sketch.update(u32::MAX);

      let mut signed_sketch = ThetaSketch::builder().build();
      signed_sketch.update(-1i64);

      assert_eq!(u32_sketch.iter().next(), signed_sketch.iter().next());
 }
```

but in c++ , follwing is success 
```
TEST_CASE("theta sketch: unsigned narrow integer inputs follow cpp
  signed path", "[theta_sketch]") {
    update_theta_sketch u32_sketch =
      update_theta_sketch::builder().build();
    u32_sketch.update(uint32_t(UINT32_MAX));

    update_theta_sketch signed_sketch =
      update_theta_sketch::builder().build();
    signed_sketch.update(int64_t(-1));

    REQUIRE(u32_sketch.get_num_retained() == 1);
    REQUIRE(signed_sketch.get_num_retained() == 1);
    REQUIRE(*u32_sketch.begin() == *signed_sketch.begin());
  }
```

because in c++, uint32 will be convert to int32_t and then int32_t. So in Rust to make it compatible with this behaviour, we introduce `SketchHashable`.
```
 void update(uint32_t value) {
    update(static_cast<int32_t>(value));
  }

  void update(int32_t value) {
    update(static_cast<int64_t>(value));
  }

  void update(int64_t value) {
    update(&value, sizeof(value));
  }
```

tisonkun (changes requested):
I get the point now.

But I think we need to consider how Rust's `Hash` would be recognized in datasketches-rust then, instead of randomly patch some code paths.

That said, we still have sketches like `BloomFilter` and `FrequentItemsSketch` that depend on Rust's `std::hash::Hash`. If we partially "fix" for CPC/HLL/Theta, then it creates a new splitting point among sketches.

To replace `std::hash::Hash` with datasketches' own `Hash` trait is one possible way to go, but then we need to consider whether this should be sealed or users can implement for their own types (technically, users can direct their own type to an u64 hashed value and then use our sealed u64 hash impl, but it may or may not desireable).

ZENOTME:
> That said, we still have sketches like BloomFilter and FrequentItemsSketch that depend on Rust's std::hash::Hash. If we partially "fix" for CPC/HLL/Theta, then it creates a new splitting point among sketches.

Yes, we should ensure that BloomFilter and FrequentItemsSketch are also consistent with the other implementations. However, I noticed that even in the C++ version, BloomFilter and Theta follow different canonicalization paths:

```
// bloom_filter
template<typename A>
void bloom_filter_alloc<A>::update(uint32_t item) {
  update(static_cast<uint64_t>(item));
}

// theta
template<typename A>
void update_theta_sketch_alloc<A>::update(uint32_t value) {
  update(static_cast<int32_t>(value));
}

template<typename A>
void update_theta_sketch_alloc<A>::update(int32_t value) {
  update(static_cast<int64_t>(value));
}
```

Instead of introducing multiple hashing behaviors using multiple type, I think it maybe more reasonable to define a single canonical hashing model and make all sketches follow it.

> To replace std::hash::Hash with datasketches' own Hash trait is one possible way to go, but then we need to consider whether this should be sealed or users can implement for their own types 

For now, I made `SketchHashable` a sealed trait to ensure behavior consistent with the Java and C++ implementations. Users can only call `update` with the supported types.

tisonkun:
Loop in @AlexanderSaydakov @leerho @proost @freakyzoidberg for general discussion on how DataSketches' sketches should handle hashable value.

Certain programming languages can have their own Hash/Hashable trait/interface and it seems the current impls vary one another.

proost:
A few thoughts:

1. Go has standard hash interfaces, but they have not been convenient for ds-go in practice. Instead of relying on a hash interface, we have generally used common utility functions where needed.
2. In ds-go, as in ds-cpp and ds-java, the hash handling seems to differ from sketch to sketch. Because of that, I had not assumed there was a single common hashing model shared across all sketches.
3. Given that, I am not yet convinced that introducing a project-wide common hash interface would provide enough benefit over using shared utility functions where appropriate.
4. A project-wide hash interface would also create a broader compatibility obligation. If a new language implementation or a new sketch is introduced later, we would need to decide whether that interface still fits all cases and whether its semantics can remain forward-compatible across the project. Most case, I think, can be covered but I'm not sure.

leerho:
As @proost points out, the different sketches use hash functions in very different ways so I'm not sure there is a common "hashing model" across all the different sketches.  

Nonetheless, in order to have cross-language compatibility each language needs to ensure that data is presented to hash functions in a deterministic and non-ambiguous way, and given the same data, the same hash algorithm in every language will result in the same hash values.

Different languages have their own primitive data representations and implicit integral and floating-point conversion rules and therefore must be dealt with on a language by language basis.  

Java was the first language that the library started with and, unfortunately, does not yet support unsigned integral primitives (except for _char_).  As a result, Java sets the common standard for the integral types.   Java supports a fairly simple set of rules for widening implicit integral conversions and floating-point conversions that are used in expressions and for method input parameters.

C++ widening implicit conversion rules are more granular and complex because it has to account for machine-dependent type sizes and unsigned variants.  As you can see in this code [snippet](https://github.com/apache/datasketches-cpp/blob/master/theta/include/theta_sketch_impl.hpp#L145-L209), a _uint8_t_ is cast to an _int8_t,_ which is then cast to an _int64_t_ before being sent to the hash function.   Although C++ can do some of this automatically, these extra methods were created to make the widening conversions explicit to ensure that data is handled correctly and consistently, leaving no doubt in the developer's mind about how the data is being handled.

I am not a Rust expert, so I'll leave it to the Rust folks to decide on the best strategy.

ZENOTME:
> Given that, I am not yet convinced that introducing a project-wide common hash interface would provide enough benefit over using shared utility functions where appropriate.

> Nonetheless, in order to have cross-language compatibility each language needs to ensure that data is presented to hash functions in a deterministic and non-ambiguous way, and given the same data, the same hash algorithm in every language will result in the same hash values.

Thanks @proost @leerho , based on the above opinion:
1. We need to align the behaviour with other languages to have cross-language compatibility.
2. It's not appropriate to introduce a project-wide common hash interface.

There are two ways to implement custom hash behaviour:
1. Introduce the hashable trait for each sketch to clean the interface. 
2. Separate the update interface like other languages, e.g., e.g update_f32, update_u32, update_i32, ...

I prefer 1 for a cleaner interface, are there other considerations? cc @tisonkun

tisonkun:
> Introduce the hashable trait for each sketch to clean the interface.

I don't think we need one for each sketch. Let me prepare a patch tomorrow.

tisonkun:
> Java was the first language that the library started with and, unfortunately, does not yet support unsigned integral primitives (except for char). As a result, Java sets the common standard for the integral types. Java supports a fairly simple set of rules for widening implicit integral conversions and floating-point conversions that are used in expressions and for method input parameters.

Unfortunately Java's `ReversePurgeItemHashMap` actually use `key.hashCode()` to calculate the hash value for items inserted into `ReversePurgeItemHashMap`.

So we are in a wild field where different sketches can have different hash algorithm on different type of value.

jmalkin:
The good news is that frequent items just uses hashing for efficient storage. It’s not fundamental to the algorithm in the same way as the distinct counting sketches (although the efficiency did still matter). Values are re-hashed on deserialization, I believe, so there’s no compatibility issue.

tisonkun:
Yes.

I have an idea to introduce CanonicalFloat, CanonicalInteger, CanonicalSlice as Hash Value adapter that should work in this case.

It's 1AM now so hopefully I can get it done tomorrow.

PR #116, opened 2026-04-13
feat(hll): expose coupon fn and improve update_with_coupon docs

notfilippo:
Make `hll::coupon` public so callers can compute a coupon from any hashable value without going through `HllSketch::update`. This enables a common optimization: pre-compute coupons for data once and insert via `update_with_coupon`, avoiding redundant hashing.

- `hll::coupon` is now `pub` with full documentation
- `HllSketch::update` cross-references `coupon` + `update_with_coupon`
- `HllSketch::update_with_coupon` gains a detailed doc comment explaining the coupon format and the multi-sketch use case

Follows up on #113 which made `update_with_coupon` public.

cc @fulmicoton-dd

### DIFF
--- a/datasketches/src/hll/mod.rs
+++ b/datasketches/src/hll/mod.rs
@@ -177,8 +177,35 @@ fn pack_coupon(slot: u32, value: u8) -> u32 {
     ((value as u32) << KEY_BITS_26) | (slot & KEY_MASK_26)
 }
 
-/// Generate a coupon from a hashable value.
-fn coupon<H: Hash>(v: H) -> u32 {
+/// Compute the HLL coupon for a hashable value.
+///
+/// A coupon is a compact 32-bit integer that encodes a bucket index (low 26 bits,
+/// derived from the low bits of a 128-bit MurmurHash) and a register value (high 6
+/// bits, derived from the leading-zero count of the high bits).  It is the internal
+/// currency of every HLL sketch.
+///
+/// Pre-computing coupons is useful when the same logical value must be inserted into
+/// multiple independent sketches, because the (relatively expensive) hash step is paid
+/// only once.  A common pattern is dictionary-encoded data: compute the coupon for each
+/// term id up front, cache it, and then call [`HllSketch::update_with_coupon`] for
+/// each per-bucket sketch rather than calling [`HllSketch::update`] repeatedly with the
+/// decoded string.
+///
+/// # Examples
+///
+/// ```
+/// # use datasketches::hll::{HllSketch, HllType, coupon};
+/// let c = coupon("hello");
+///
+/// let mut sketch1 = HllSketch::new(10, HllType::Hll8);
+/// let mut sketch2 = HllSketch::new(12, HllType::Hll8);
+/// sketch1.update_with_coupon(c);
+/// sketch2.update_with_coupon(c);
+///
+/// assert!(sketch1.estimate() >= 1.0);
+/// assert!(sketch2.estimate() >= 1.0);
+/// ```
+pub fn coupon<H: Hash>(v: H) -> u32 {
     let mut hasher = MurmurHash3X64128::default();
     v.hash(&mut hasher);
     let (lo, hi) = hasher.finish128();
--- a/datasketches/src/hll/mod.rs
+++ b/datasketches/src/hll/mod.rs
@@ -157,32 +157,12 @@ const COUPON_RSE: f64 = COUPON_RSE_FACTOR / (1 << 13) as f64;
 const RESIZE_NUMERATOR: u32 = 3; // Resize at 3/4 = 75% load factor
 const RESIZE_DENOMINATOR: u32 = 4;
 
-/// Extract slot number (low 26 bits) from coupon
-#[inline]
-fn get_slot(coupon: u32) -> u32 {
-    coupon & KEY_MASK_26
-}
-
-/// Extract value (upper 6 bits) from coupon
-#[inline]
-fn get_value(coupon: u32) -> u8 {
-    (coupon >> KEY_BITS_26) as u8
-}
-
-/// Pack slot number and value into a coupon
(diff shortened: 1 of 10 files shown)


notfilippo:
cc @jmalkin . Open for discussion if we should remove the feature entirely (drop `pub` in `update_with_coupon`) or embrace this as a feature.

fulmicoton:
To bullet proof this, should we introduce a Coupon struct?

```rust
/// A coupon encodes a (slot, value) pair derived from hashing an input.
///
/// Format: `[value (6 bits) << 26] | [slot (26 bits)]`
///
/// The slot identifies an HLL register (derived from the lower bits of the hash),
/// and the value represents the number of leading zeros plus one (from the upper bits).
#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
pub struct Coupon(u32);

impl Coupon {
    /// Sentinel value indicating an empty coupon slot.
    const EMPTY: Self = Coupon(0);

    /// Create a coupon by hashing a value.
    #[inline(always)]
    pub fn from_hash(v: impl Hash) -> Self {
        let mut hasher = MurmurHash3X64128::default();
        v.hash(&mut hasher);
        let (lo, hi) = hasher.finish128();

        let addr26 = lo as u32 & KEY_MASK_26;
        let lz = hi.leading_zeros();
        let capped = lz.min(62);
        let value = capped + 1;

        Coupon((value << KEY_BITS_26) | addr26)
    }

    /// Pack slot number and value into a coupon.
    ///
    /// Format: [value (6 bits) << 26] | [slot (26 bits)]
    #[inline(always)]
    fn pack(slot: u32, value: u8) -> Self {
        Coupon(((value as u32) << KEY_BITS_26) | (slot & KEY_MASK_26))
    }

    /// Extract slot number (low 26 bits) from coupon.
    #[inline(always)]
    fn slot(self) -> u32 {
        self.0 & KEY_MASK_26
    }

    /// Extract value (upper 6 bits) from coupon.
    #[inline(always)]
    fn value(self) -> u8 {
        (self.0 >> KEY_BITS_26) as u8
    }
}
```

(a commit that does this https://github.com/fulmicoton-dd/datasketches-rust/commit/6d65e11278ea41182e079485c143385da44bd245)

notfilippo:
I really like this approach @fulmicoton. Can I steal the commit from you and include it in this PR?

fulmicoton:
Absolutely but I did not work much over it and could have forgotten places

datasketches/src/hll/mod.rs
```diff
@@ -157,52 +157,102 @@ const COUPON_RSE: f64 = COUPON_RSE_FACTOR / (1 << 13) as f64;
...
+pub struct Coupon(u32);
 
-/// Extract value (upper 6 bits) from coupon
-#[inline]
-fn get_value(coupon: u32) -> u8 {
-    (coupon >> KEY_BITS_26) as u8
-}
+impl Coupon {
+    /// Sentinel value indicating an empty coupon slot.
+    const EMPTY: Self = Coupon(0);
```
fulmicoton, line 193: I think it is worth making EMPTY pub (if possible)

```suggestion
    pub const EMPTY: Self = Coupon(0);
```

tisonkun:
This can be related to https://github.com/apache/datasketches-rust/pull/111 that at which level we integrate with Rust's `std::hash::Hash` trait.

tisonkun:
That is, perhaps we can always accept only a hashed value on sketches, and provide an adapter to calculate the hashed value for struct that implements `std::hash::Hash` or other well-known strategy/algorithm.

cc @ZENOTME 

So `Coupon::from_hash` use Rust's native Hash impl, while we provide `Coupon::from_hash_compat` to be compatible with other DataSketches impls like datasketches-cpp.

datasketches/src/hll/mod.rs
notfilippo: I didn't expose it because inserting it in the sketch would break the marker logic.

notfilippo:
> So Coupon::from_hash use Rust's native Hash impl, while we provide Coupon::from_hash_compat to be compatible with other DataSketches impls like datasketches-cpp.

Make sense to me, we can introduce it in a later PR.

datasketches/src/hll/mod.rs
fulmicoton-dd: I see!

ZENOTME:
> That is, perhaps we can always accept only a hashed value on sketches, and provide an adapter to calculate the hashed value for struct that implements `std::hash::Hash` or other well-known strategy/algorithm.
> 
> cc @ZENOTME
> 
> So `Coupon::from_hash` use Rust's native Hash impl, while we provide `Coupon::from_hash_compat` to be compatible with other DataSketches impls like datasketches-cpp.

Make sense to me

tisonkun (approved):
LGTM.

Hopefully I can have some cycles the next week to give an overall review of the hashed value design.

jmalkin:
I don’t get the different hash variants for compatibility? All the hash values must be an exact bit match across all languages or the sketches are not compatible.

tisonkun:
~@notfilippo I noticed that the only pub constructor of `Coupon` is `Coupon::from_hash`~
~Then could you show me a real-world use case how making `Coupon` pub different from:~

UPDATE: I got it now you can compute the hashed value once.

```rust
let mut sketch = HllSketch::new(10, HllType::Hll8);
sketch.update_with_coupon(Coupon::from_hash("apple"));
// the same as sketch.update("apple");
```

> I don’t get the different hash variants for compatibility? All the hash values must be an exact bit match across all languages or the sketches are not compatible.

Yes. All hashed values are exact the same. The issue here is how we hash values.

For example, in Java, there is no `uint_8` and `uint_16`, so it does not need to consider how to hash `uint_8` (`u8` in rust) when users insert it.

PR #117, opened 2026-04-13
fix(hll): update silently dropped after deserializing a compact List sketch

### LINKED ISSUE #115: bug upgrading a deserialized hll sketch
When serializing/deserializing a sketch we create a "compact" version of the datasketch.
For List, this means the array of values is as long as the len of the lists.

The update code on the end, attempts to append the value by scanning through the list of values, 
up to the point where a SENTINEL value is reached (value 0) and inserts it there.

Due to the compact representation of lists, we are never able to add an element.

If the end of the list is reached, we do not panic, and exit silently.
After updating, we see if we need to upgrade the list to a set.

Minimal reproducing test.

```rust

    #[test]
    fn hll_serialize_deserialize_then_update() {
        const LG_K: u8 = 11;
        let mut sketch = HllSketch::new(LG_K, HllType::Hll4);

        sketch.update(1);

        // Serialize / Deserialize round-trip
        let bytes = sketch.serialize();
        let mut sketch = HllSketch::deserialize(&bytes).unwrap();
        sketch.update(2);

        // The sketch should have observed 3 distinct coupons
        let est = sketch.estimate();
        assert!(
            (est - 2.0).abs() < 0.1,
            "expected estimate close to 2, got {est}"
        );
    }
```

notfilippo:
When an `HllSketch` in **List mode** is serialized it uses compact format: only the live coupons are written to disk, with no trailing `COUPON_EMPTY` sentinels. On deserialization, the container was allocated with exactly `coupon_count` slots, leaving the array fully packed.

`list::update()` scans linearly for a `COUPON_EMPTY` (`0`) sentinel to find an insertion slot. Finding none, it fell through silently — discarding every value added after the round-trip.

Always allocate `1 << lg_arr` slots (initialized to `COUPON_EMPTY`) in `List::deserialize()`, reading only `coupon_count` elements from the byte stream in compact mode. The trailing empty slots are then available as sentinels for subsequent `update()` calls.

Fixes #115.

### DIFF
--- a/datasketches/src/hll/list.rs
+++ b/datasketches/src/hll/list.rs
@@ -82,13 +82,17 @@ impl List {
         empty: bool,
         compact: bool,
     ) -> Result<Self, Error> {
-        // Compute array size
-        let array_size = if compact { coupon_count } else { 1 << lg_arr };
+        // Always allocate the full-sized array (1 << lg_arr) so Coupon::EMPTY sentinel
+        // slots are available for future update() calls. In compact format only
+        // coupon_count values are stored on disk, but memory must hold the full capacity
+        // so the linear scan in update() can find an empty slot to insert into.
+        let array_size = 1 << lg_arr;
+        let read_count = if compact { coupon_count } else { array_size };
 
-        // Read coupons
+        // Read coupons into the front of the full-sized array; remaining slots stay Coupon::EMPTY.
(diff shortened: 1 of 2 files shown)


notfilippo:
cc @tisonkun @fulmicoton

tisonkun (approved):
Thanks for the fix!

Issue #109, opened 2026-03-22
Release 0.3.0


tisonkun:
I'm going to start the preparing and voting to datasketches-rust 0.3.0 with what we have now. Other ongoing initiatives will be tracked down later, and I'd start an epic issue for the next iteration.

@PsiACE @ZENOTME @notfilippo any suggestions on releasing 0.3.0? The current Unreleased changelog is like:

```md
## Unreleased

### Breaking changes

* `CountMinSketch` now has a type parameter for the count type. Possible values are `u8` to `u64` and `i8` to `i64`.
* `HllUnion::get_result` is renamed to `HllUnion::to_sketch`.

### New features

* `CountMinSketch` with unsigned values now supports `halve` and `decay` operations.
* `CpcSketch` and `CpcUnion` are now available for cardinality estimation.
* `CpcWrapper` is now available for reading estimation from a serialized CpcSketch without full deserialization.
* `FrequentItemsSketch` now supports serde for any value implement `FrequentItemValue` (builtin supports for `i64`, `u64`, and `String`).
* Expose `codec::SketchBytes`, `codec::SketchSlice`, and `FrequentItemValue` as public API.
```

cc @leerho

ZENOTME:
Should we include following? 
- `CompactThetaSketch` is avaliable 
- ThetaIntersection operation is avaliable

tisonkun:
@ZENOTME I'd be glad to review any patch if it were available before April. Or else I may not wait for too long.

PsiACE:
I have no comments; let’s move forward with the release of 0.3.0.

leerho:
LGTM

notfilippo:
Maybe we should include the fix for #115 .

notfilippo:
@tisonkun on my side I think we can move forward with the release.

tisonkun:
Yeah. I'd like to see if we can have a consensus (conclusion and perhaps solution) on https://github.com/apache/datasketches-rust/pull/111#issuecomment-4229345925 or else we'd introduce a breaking change immediately after the release.

Perhaps I'd finish it this week or we can anyway release 0.3.0 and leave it for 0.4.0.

tisonkun:
All preparations have been done.

I'd cut the release candidate and start the vote tomorrow or the day after.

notfilippo:
Thanks for the preparation @tisonkun :)

Could we also land this fix for HLL in the release? #122

tisonkun:
[...]

PR #118, opened 2026-05-07
refactor: introduce hash_value for compatible hashing

tisonkun:
This closes #111

This refers to #109 

cc @notfilippo @ZENOTME @PsiACE @Xuanwo 

cc @jmalkin @leerho @proost


PsiACE (approved):
lgtm

datasketches/src/hash_value/canonical.rs
```diff
@@ -0,0 +1,515 @@
...
+/// assert_eq!(
+///     calculate_hash(canonical_u32(42)),
+///     calculate_hash(canonical_i64(42))
+/// );
+/// assert_ne!(
+///     calculate_hash(canonical_u32(2147483648)),
+///     calculate_hash(canonical_i64(2147483648))
+/// );
+/// ```
+pub fn canonical_u32(v: u32) -> Canonical<u32> {
```
ZENOTME: https://github.com/apache/datasketches-rust/pull/111#issuecomment-4229316731 bloom_filter have different canonical_u32 behaviour. How we deal with that in the future🤔 
```
// bloom_filter
template<typename A>
void bloom_filter_alloc<A>::update(uint32_t item) {
  update(static_cast<uint64_t>(item));
}
```

ZENOTME (approved):
Other LGTM.

datasketches/src/hash_value/canonical.rs
tisonkun: It's something about naming.

We can for sure have a new value wrapper for this manner. Let me restructure the code.

tisonkun: ... and damn it datasketches-cpp has different integer extension strategy for different sketches 🤣 

I like the zero extension but HLL/CPC with sign extension anyway.

tisonkun: Resolved.

datasketches/src/hash_value/mod.rs
```diff
@@ -0,0 +1,91 @@
...
+//! Empty byte and string inputs have zero bytes to hash. Other datasketches implementations skip
+//! empty strings before hashing, so check `is_empty` before updating a sketch when that behavior
+//! matters.
+//!
+//! Read the docs of concrete value wrapper for more details and examples.
+//!
+//! * [`raw_bytes::from_vec`]
+//! * [`raw_bytes::from_string`]
+//! * [`raw_bytes::from_slice`]
+//! * [`raw_bytes::from_str`]
```
tisonkun, line 73: @proost @freakyzoidberg I'd like to know whether datasketches-go follows the same hashing strategy for the related sketches.

datasketches/src/hash_value/mod.rs
```diff
@@ -0,0 +1,91 @@
...
+//!
+//! Read the docs of concrete value wrapper for more details and examples.
+//!
+//! * [`sign_extend::from_i8`], [`sign_extend::from_u8`]
+//! * [`sign_extend::from_i16`], [`sign_extend::from_u16`]
+//! * [`sign_extend::from_i32`], [`sign_extend::from_u32`]
+//!
+//! [`natural_extend::NaturalExtend`] first widens signed values to `i64` and unsigned values to
+//! `u64`, and then hashes the resulting integers. This strategy is the same as how datasketches-cpp
+//! hashes short integers for `HllSketch` and `CpcSketch`.
```
ZENOTME: ~~hashes short integers for `HllSketch` and `CpcSketch`.~~

hashes short integers for `BloomFilter`.

tisonkun: Updated.

tisonkun:
Merging ...

Open for review after merge.

I'll try to gate sketches with feature flags and then cut a release candidate.

datasketches/src/cpc/sketch.rs
```diff
@@ -172,7 +171,25 @@ impl CpcSketch {
...
+    /// sketch.update(1);
+    /// sketch.update(2);
+    /// sketch.update(3);
+    ///
+    /// let mut sketch = CpcSketch::with_seed(11, 123);
+    /// sketch.update(hash_value::canonical_float::from_f64(1.5));
+    /// sketch.update(hash_value::canonical_float::from_f64(2.5));
+    /// sketch.update(hash_value::canonical_float::from_f64(3.5));
+    /// ```
     pub fn update<T: Hash>(&mut self, value: T) {
```
tisonkun, line 193: One possible future approach to prevent users from accidentally using the Rust standard library's Hash trait without realizing it, is to implement our own `HashValue` or `Hashable` trait.

We could then provide a simple wrapper for Rust's std Hash like our current hash_value wrappers. In this scenario, users would be required to either use our wrapper when passing data to `update`, or implement the `Hashable` logic themselves.

I think this could serve as a potential future extension. However, forcing users to use a specific wrapper or an additional `Hashable` trait might feel unnatural at the interface level, especially since we would never have a blanket `impl<T: Hash> Hashable for T`.

I will leave this here as a record. If we find a need for this in the future, we can revisit and implement this idea.

FYI @ZENOTME @notfilippo

PR #120, opened 2026-05-09
refactor: gate sketches with feature flags 

### LINKED ISSUE #32: Create a Cargo feature flag for each sketch
Motivation:
1. We can gate compiler warnings only when certain sketches are tying to be compiled on unsupported platforms (e.g. #28)
2. We can allow users to pick and choose their sketches without having to compile the whole crate

By default, all sketches should be turned on IMO.

tisonkun:
This closes #32 

Some unused warnings and style issues can be fixed later.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -35,7 +35,6 @@ datasketches = { path = "datasketches" }
 clap = { version = "4.5.20", features = ["derive"] }
 insta = { version = "1.46.1" }
 googletest = { version = "0.14.2" }
-rand = { version = "0.9.2" }
 which = { version = "8.0.0" }
 
 [workspace.lints.rust]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,7 +25,7 @@ homepage = "https://datasketches.apache.org"
 license = "Apache-2.0"
 readme = "README.md"
 repository = "https://github.com/apache/datasketches-rust"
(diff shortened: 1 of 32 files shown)


Cargo.toml
```diff
@@ -35,7 +35,7 @@ datasketches = { path = "datasketches" }
 clap = { version = "4.5.20", features = ["derive"] }
 insta = { version = "1.46.1" }
 googletest = { version = "0.14.2" }
-rand = { version = "0.9.2" }
+cargo_metadata = { version = "0.19.1" }
```
tisonkun: Latest version 0.23.1 requires 1.86 and the main branch bumps MSRV to 1.88.

I think we can anyway consider bump MSRV later.

tisonkun: Update MSRV to 1.86.0 to simplify workflow.

PsiACE (approved):

tisonkun:
Thanks for your review @PsiACE!

I'm going to merge this and cut a 0.3.0 release candidate in the next Monday. So I'd appreciate it if @notfilippo @ZENOTME can drop a review so that we don't need to go back for hotfix after the RC is out.

tisonkun:
Well. I think I can merge this now and prepare the changelog and other setups. And feedback can steer in to break the release process :D

PR #122, opened 2026-05-11
fix: correct assetion logic

tabac:
For exceptions stored in the auxiliary map, the corresponding register should be the max possible value (`AUX_TOKEN`).

I believe we have to assert the opposite of what we do in that case, that is that the value of the slot is `AUX_TOKEN`.

### DIFF
--- a/datasketches/src/hll/array4.rs
+++ b/datasketches/src/hll/array4.rs
@@ -230,10 +230,10 @@ impl Array4 {
             let mut new_aux = None;
 
             for (slot, old_actual_val) in old_aux.into_iter() {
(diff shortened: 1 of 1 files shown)


notfilippo (approved):
Makes sense to me! Thanks for catching

tisonkun (approved):
LGTM. This should be guarded with a unit test, which we can add later.

tisonkun:
Thanks for your contribution @tabac! I added a regression test at https://github.com/apache/datasketches-rust/pull/123. Welcome to drop a review there.

PR #128, opened 2026-05-13
fix(hll): preserve set coupon order when serializing

notfilippo:
Compact HLL Set serialization used to sort retained coupons before writing bytes. Java and C++ write compact Set coupons in hash table iteration order.

Write the existing iterator output directly. This keeps the compact format and removes the extra allocation.

### DIFF
--- a/datasketches/tests/hll_serialization_test.rs
+++ b/datasketches/tests/hll_serialization_test.rs
@@ -23,6 +23,7 @@ use std::fs;
 use std::path::PathBuf;
 
 use common::serialization_test_data;
+use datasketches::hash_value::natural_extend;
 use datasketches::hll::HllSketch;
 use datasketches::hll::HllType;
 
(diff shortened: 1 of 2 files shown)


tisonkun:
Do we have a test to capture this issue? I can't simply understand the situation by barely reading this patch.

notfilippo:
> Do we have a test to capture this issue? I can't simply understand the situation by barely reading this patch.

I am working on a test but you were too fast 🏎️ ! Will mark as draft and work on it :)

tisonkun:
> > Do we have a test to capture this issue? I can't simply understand the situation by barely reading this patch.
> 
> I am working on a test but you were too fast 🏎️ ! Will mark as draft and work on it :)

Great!

I wonder if you'd suggest to cancel 0.3.0-rc.2 for this fix, or we may later release a (quick) 0.3.1 with this fix.

notfilippo:
It's very low priority fix which does not break compatibility. IMO we can release it in > 0.3.0.

Merging this patch would just save an allocation and make it bit-perfect with other libraries when the insertion order is identical.

notfilippo:
Added the test. Not amazingly happy with it but I think it's ok. I might add some comments later. Will mark ready for review after I'm done thinking about it :)

datasketches/tests/hll_serialization_test.rs
```diff
@@ -132,6 +133,61 @@ fn test_update_after_deserialize_list_mode() {
...
+            2 => "HLL",
+            _ => "unknown",
+        }
+    }
+
+    for (hll_type, type_name) in [
+        (HllType::Hll4, "hll4"),
+        (HllType::Hll6, "hll6"),
+        (HllType::Hll8, "hll8"),
+    ] {
```
tisonkun, line 152: nit: perhaps use `"{hll_type:?}"` and reduce the `type_name` manual mapping.

tisonkun (commented):
Verified locally that the test case failed before the impl change.

Generally LGTM.

datasketches/tests/hll_serialization_test.rs
notfilippo: Makes sense but also it's needed to load serialized sketch files that have "hllX" in the name

tisonkun: Oops. That's it.

notfilippo:
@tisonkun could you re-approve? Thanks!

tisonkun (approved):
LGTM. Thanks!

PR #131, opened 2026-05-18
chore: Update CHANGELOG for v0.3.0

tisonkun:
Document breaking changes for version 0.3.0.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@ All significant changes to this project will be documented in this file.
 
 ## Unreleased
 
+## v0.3.0 (2026-05-18)
+
 ### Breaking changes
 
 * `CountMinSketch` now has a type parameter for the count type. Possible values are `u8` to `u64` and `i8` to `i64`.


CHANGELOG.md
```diff
@@ -4,6 +4,8 @@ All significant changes to this project will be documented in this file.
 
 ## Unreleased
```
notfilippo, line 5: We can drop this "Unreleased"?

notfilippo (approved):

CHANGELOG.md
tisonkun: Nope. It's a stub for the next unreleased changelog. An empty "Unrelesed" section would not cause misunderstanding and it simplifies contributors to attatch changelog entries instead of the very first contributor should add an "Unreleased" heading.

notfilippo: Maybe it would be nice to add an HTML style comment to denote that. Thanks for the explanation.

tisonkun: Would you submit a PR to add one?

It's intuitive to me when I learn this pattern from somewhere like https://github.com/open-telemetry/opentelemetry-rust/blob/b1393e31c12543e704be6d4595b3d22134c59f9b/opentelemetry-http/CHANGELOG.md.

tisonkun: I adopt this pattern for most of my projects, BTW (e.g., https://github.com/scopedb/scopeql/blob/2e34a2cd47f6996874e45b1ddac030f643b2abe4/CHANGELOG.md)

notfilippo: > Would you submit a PR to add one?

Sure. Why not! Looks actually like a fairly nice approach and might enforce PR creators to also include an entry in the CHANGELOG.md so it can always be up-to-date with unreleased changes.

PR #132, opened 2026-05-19
Bugfix: HLL4 written in rust cannot not be loaded in Java

fulmicoton-dd:
The aux map can be serialized in two format.
Rust serializes only the compact format but omits to set the compact flag.

https://github.com/apache/datasketches-java/blob/main/src/main/java/org/apache/datasketches/hll/HeapAuxHashMap.java#L65-L96

This PR sets the flag.

### DIFF
--- a/datasketches/src/hll/array4.rs
+++ b/datasketches/src/hll/array4.rs
@@ -29,6 +29,7 @@ use crate::common::NumStdDev;
 use crate::error::Error;
 use crate::hll::Coupon;
 use crate::hll::estimator::HipEstimator;
+use crate::hll::serialization::COMPACT_FLAG_MASK;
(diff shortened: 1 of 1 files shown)


notfilippo (approved):
Looks good to me.

@tisonkun , should we look into testing that our serialized sketches can be read correctly the Java / C++ implementation?

tisonkun:
@notfilippo I think we can file an issue on both repo and collaborate with perhaps @leerho and @AlexanderSaydakov to land the compatibility tests.

It can be also related to our long-term idea on #10

tisonkun:
@notfilippo side note - we may use `fix: ..` to try to follow the semantic commit convention always.

PR #138, opened 2026-06-23
feat: add tuple theta implementation and tests

ariesdevil:
## Summary

## Note

### DIFF
--- a/datasketches/src/tuple/hash_table.rs
+++ b/datasketches/src/tuple/hash_table.rs
@@ -17,20 +17,12 @@
 
 use std::hash::Hash;
 
-use crate::common::ResizeFactor;
-use crate::hash::MurmurHash3X64128;
-use crate::hash::compute_seed_hash;
-use crate::theta::HASH_TABLE_REBUILD_THRESHOLD;
-use crate::theta::HASH_TABLE_RESIZE_THRESHOLD;
-use crate::theta::MAX_THETA;
-use crate::theta::MIN_LG_K;
-use crate::theta::STRIDE_MASK;
-use crate::theta::starting_sub_multiple;
-use crate::theta::starting_theta_from_sampling_probability;
+use crate::theta::RawHashTable;
+use crate::theta::RawHashTableEntry;
 
 /// A retained entry: a hash key together with its associated summary.
 #[derive(Debug)]
-struct TupleEntry<S> {
+pub(super) struct TupleEntry<S> {
     hash: u64,
     summary: S,
 }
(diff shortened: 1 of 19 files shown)


ariesdevil:
@tisonkun @PsiACE PTAL if you have time.

tisonkun:
Thanks for your contribution @ariesdevil! I'll review this in the next week.

datasketches/src/tuple/hash_table.rs
```diff
@@ -0,0 +1,780 @@
...
+/// summary. It maintains an array with capacity up to 2^lg_max_size:
+/// * Before it reaches the max capacity, it will extend the array based on resize_factor.
+/// * After it reaches the capacity bigger than 2^lg_nom_size, every time the number of entries
+///   exceeds the threshold, it will rebuild the table: only keep the min 2^lg_nom_size entries and
+///   update the theta to the k-th smallest entry.
+///
+/// Unlike the Theta hash table, when a key is inserted that already exists, the incoming update is
+/// merged into the existing summary rather than discarded.
+#[derive(Debug)]
+pub(super) struct TupleHashTable<S> {
```
ZENOTME: Since this hash table is same as `ThetaHashTable`. Like cpp, maybe we should refactor it to use Generics type and avoid duplicate implementation.

ZENOTME:
Based on https://github.com/apache/datasketches-rust/pull/138#discussion_r3530833211 , looks like it's better to refactor ThetaHashTable, related set operation to generic version first. So that we can reuse them in tuple implementation.

ariesdevil:
Fixed, PTAL again.

datasketches/src/tuple/union.rs
```diff
@@ -0,0 +1,495 @@
...
+///
+/// let mut union = TupleUnion::<u64>::builder().build();
+/// union.update(&a).unwrap();
+/// union.update(&b).unwrap();
+///
+/// let result = union.result(true);
+/// assert_eq!(result.num_retained(), 3); // apple, banana, cherry
+/// ```
+#[derive(Debug)]
+pub struct TupleUnion<S, P = DefaultUnionPolicy> {
```
ZENOTME, line 65: This also can be `ThetaUnion<E, P = DefaultUnionPolicy>` and reuse between tuple and theta.

ZENOTME:
Thanks @ariesdevil ! This PR contains several part. How about separate this PR into smaller so that it can be more easier to review e.g. 
- theta refactor + tuple sketch 
- set op(union, intersection, ...)

datasketches/src/tuple/union.rs
ariesdevil: Seems theta has no union implementation.

ZENOTME: Yes. I send a implementation : https://github.com/apache/datasketches-rust/pull/145 . But it's a intermediate state for theta. I think in final we should have something like `ThetaUnion<E, P = DefaultUnionPolicy>` (or `RawThetaUnion<E, P = DefaultUnionPolicy>`) so that tuple union and theta union can reuse it in both.

In https://github.com/apache/datasketches-rust/pull/145 I didn't implement it in this way is because we need to have RawHashTable first. 🤣

ariesdevil: So we should merge #145 then I'll rebase on it.

ZENOTME: We can separate out `RawHashTable` as a PR first. After merge it, #145 can be rewrite as generic type version. And then tuple union can reuse it.

ariesdevil: Done at #146

PR #140, opened 2026-06-25
feat(frequencies): add borrowed-key update methods to the frequent items sketch

k3dom:
> **Disclaimer:** This contribution was written entirely by an AI coding
> assistant (Claude Opus 4.8) and has only been manually reviewed.

## Motivation

`FrequentItemsSketch::update` and `update_with_count` take the item by value
(`T`). When `T` is an owned, heap-allocated type such as `String`, a caller that
streams many repeated items must allocate an owned key for *every* observation —
including when the item is already tracked, in which case
`adjust_or_put_value` only compares the key by reference and then immediately
drops the freshly allocated value. For a stream dominated by a small set of
recurring items (a common case for a frequent-items sketch), that is one wasted
allocation per occurrence.

## Change

Add borrowed-key update methods that allocate an owned key only when the item is
actually inserted:

- `FrequentItemsSketch::update_borrowed<Q>(&Q)`
- `FrequentItemsSketch::update_with_count_borrowed<Q>(&Q, u64)`
- `ReversePurgeItemHashMap::adjust_or_put_value_borrowed<Q>(&Q, u64)` (backs the
  two above)

They mirror the existing `update` / `update_with_count` pair and are bounded by
`T: Borrow<Q>, Q: Eq + Hash + ToOwned<Owned = T> + ?Sized`, so for example a
`FrequentItemsSketch<String>` can be updated directly from a `&str`. The hit
path increments the counter with no allocation; the insert path calls
`key.to_owned()` exactly as the owned path does today. `hash_item` is relaxed to
`T: Hash + ?Sized` so unsized borrowed keys such as `str` can be hashed
directly.

## Testing

- Added doctests for both new sketch methods.
- `cargo test -p datasketches --features frequencies --lib` and `--doc` pass
  (including the new doctests).

### DIFF
--- a/datasketches/src/frequencies/reverse_purge_item_hash_map.rs
+++ b/datasketches/src/frequencies/reverse_purge_item_hash_map.rs
@@ -68,7 +68,7 @@ impl<T: Eq + Hash> ReversePurgeItemHashMap<T> {
         T: Borrow<Q>,
         Q: Eq + Hash + ?Sized,
     {
-        let probe = self.hash_probe(key);
+        let (probe, _) = self.find_probe_or_empty(key);
         if self.states[probe] > 0 {
             return self.values[probe];
         }
@@ -77,21 +77,7 @@ impl<T: Eq + Hash> ReversePurgeItemHashMap<T> {
 
     /// Adds `adjust_amount` to the value for `key`, inserting if absent.
     pub fn adjust_or_put_value(&mut self, key: T, adjust_amount: u64) {
-        let mask = self.keys.len() - 1;
-        let mut probe = (hash_item(&key) as usize) & mask;
-        let mut drift: usize = 1;
-        while self.states[probe] != 0 {
-            let matches = self.keys[probe]
-                .as_ref()
-                .map(|existing| existing == &key)
-                .unwrap_or(false);
-            if matches {
-                break;
-            }
-            probe = (probe + 1) & mask;
-            drift += 1;
-            debug_assert!(drift < DRIFT_LIMIT, "drift limit exceeded");
-        }
+        let (probe, drift) = self.find_probe_or_empty(&key);
         if self.states[probe] == 0 {
             self.keys[probe] = Some(key);
             self.values[probe] = adjust_amount;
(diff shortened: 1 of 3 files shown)


tisonkun (commented):
We are still in a 0.x series, so breaking changes can happen.

I suppose a refactor on `item` parameter to use `K: Borrow<Q>` always, like what `HashMap` does, would make the interface clear and concise to use.

What do you think?

tisonkun (changes requested):
After some local tests, I think we can have:

```rust
    pub fn estimate<Q>(&self, item: &Q) -> u64
    where
        T: Borrow<Q>,
        Q: Hash + Eq + ?Sized,
    {
        let value = self.hash_map.get(item);
        if value > 0 { value + self.offset } else { 0 }
    }

    pub fn lower_bound<Q>(&self, item: &Q) -> u64
    where
        T: Borrow<Q>,
        Q: Hash + Eq + ?Sized,
    {
        self.hash_map.get(item)
    }

    pub fn upper_bound<Q>(&self, item: &Q) -> u64
    where
        T: Borrow<Q>,
        Q: Hash + Eq + ?Sized,
    {
        self.hash_map.get(item) + self.offset
    }
```

But for passing-in owned value for `update`/`update_with_count`, even `HashMap::insert` requires:

```rust
    pub fn insert(&mut self, k: K, v: V) -> Option<V> {
        self.base.insert(k, v)
    }
```

That is, the caller is responsible to create the owned value of type `K`. Trying to provide convenient interfaces for arbitrary types, said `Q: Eq + Hash + ToOwned<Owned = T> + ?Sized` would be random and introduce more confusion.

tisonkun:
If you'd like to simplify between `String` and `&str` and don't care about the clone, you can have a caller-side helper function:

```rust
fn update_items(sketch: FrequentItemsSketch<String>, k: impl Into<String>) {
  let k = k.into();
  sketch.update(k);
}
```

k3dom:
Agree on making estimate/lower_bound/upper_bound generic over `Borrow<Q>` and `ToOwned<Owned = T>` being an odd bound, will change.

On update though, insert always stores the key so it has to own it, but `adjust_or_put_value` is get-or-increment: when the item's already tracked (the common case here) it just bumps a counter and never stores the key, so there's nothing to own. Closer to the entry API than insert. std's `entry` can't take a borrowed key either, which is why e.g. hashbrown has `entry_ref`. Otherwise a hot-key stream allocates a `String` per hit just to drop it.

tisonkun:
> `entry_ref`

https://docs.rs/hashbrown/latest/hashbrown/struct.HashMap.html#method.entry_ref

This looks like a good interface to follow. You may follow its naming and trait-bound conventions, and I'll give a review once you get all updates ready.

I'm fine with ToOwned now. But the API name needs some considerations, `update_borrowed` doesn't sound like a good name.

k3dom:
Ok I followed hashmaps lead and went with a `_ref` suffix instead of the `_borrowed` what do you think? Also reverted to `ToOwned` again same as hashmaps `entry_ref`

tisonkun:
@k3dom Thanks for your feedback! Generally in a good direction.

I'm going to review this in the following week and get it merge.

This impl is OK now but I need a bit more time to consider the name and internal impl structure.

tisonkun (approved):
LGTM. Pushed a new commit for code tidy.

Let's move forward with the current name. I found it the best one I can imagine now :D

PR #143, opened 2026-07-03
fix(frequencies): write the full empty preamble long in serialize

k3dom:
## What

`FrequentItemsSketch::serialize` emits only 6 bytes for an empty sketch: the preamble declares one 8-byte preamble long, but serialization stops after the flags byte. `deserialize` consumes the full 8-byte preamble before checking the empty flag, so a serialized empty sketch always fails to deserialize with `insufficient data: <unused>`. The truncated form also diverges from the Java and C++ implementations, which both write the full preamble long for empty sketches.

Empty sketches are reachable from non-trivial streams: a map saturated with count-1 items (e.g. unique identifiers) purges every counter, leaving an empty sketch that then cannot be persisted and restored.

## Changes

- Write the two unused padding bytes in the empty path of `serialize_inner`, making the empty encoding one full preamble long and byte-compatible with the Java/C++ readers.
- Add round-trip tests for a fresh empty sketch (pinning the 8-byte length) and for a sketch purged to empty.
- Add a changelog entry under Unreleased.


tisonkun (approved):
Thanks!

PR #144, opened 2026-07-03
perf: avoid clone on serialize frequency items

tisonkun:
cc @k3dom

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ All significant changes to this project will be documented in this file.
 ### New features
 
 * `FrequentItemsSketch` now supports borrowed-key updates via `update_ref` and `update_with_count_ref`, allowing sketches such as `FrequentItemsSketch<String>` to update from `&str` without allocating on existing-key hits. Frequency queries also accept borrowed key forms matching `Borrow<Q>`.
+* `FrequentItemsSketch` no longer requires item types to implement `Clone` for core updates, queries, and serialization. Custom `FrequentItemValue` implementations can now be non-`Clone`; APIs that return or merge owned items still require `Clone`.
 
 ### Bug fixes
 
--- a/datasketches/src/frequencies/reverse_purge_item_hash_map.rs
+++ b/datasketches/src/frequencies/reverse_purge_item_hash_map.rs
@@ -197,37 +197,17 @@ impl<T: Eq + Hash> ReversePurgeItemHashMap<T> {
         self.num_active
     }
 
-    /// Returns the active keys in the map.
-    pub fn active_keys(&self) -> Vec<T>
-    where
-        T: Clone,
-    {
-        if self.num_active == 0 {
-            return vec![];
-        }
-        let mut keys = Vec::with_capacity(self.num_active);
+    /// Returns active keys and values in storage order.
+    pub fn active_entries(&self) -> Vec<(&T, u64)> {
+        let mut entries = Vec::with_capacity(self.num_active);
         for i in 0..self.keys.len() {
             if self.states[i] > 0 {
                 if let Some(key) = self.keys[i].as_ref() {
-                    keys.push(key.clone());
+                    entries.push((key, self.values[i]));
                 }
             }
         }
-        keys
-    }
-
-    /// Returns the active values in the map.
-    pub fn active_values(&self) -> Vec<u64> {
-        if self.num_active == 0 {
-            return vec![];
-        }
-        let mut values = Vec::with_capacity(self.num_active);
-        for i in 0..self.values.len() {
-            if self.states[i] > 0 {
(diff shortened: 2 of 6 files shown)


k3dom:
Nice, LGTM!

Two small thoughts:
- `active_entries()` could probably be a lazy iterator `impl Iterator<Item = (&T, u64)> + '_` instead of collecting `Vec<(&T, u64)>` which would make it totally allocation-free, not sure if it would be always faster though for very large maps.
- Would probably also make sense to update `frequent_items_with_threshold` as a follow-up because here we still clone every qualifying item into owned `Row`'s before the caller can filter

tisonkun:
@k3dom 

> active_entries() could probably be a lazy iterator impl Iterator<Item = (&T, u64)> + '_ instead of collecting Vec<(&T, u64)> which would make it totally allocation-free, not sure if it would be always faster though for very large maps.

This is proposed by my Codex for the first time. But I reject it because an `active_entries` iterator would do the activity check every time, while we'd iterate it several (now 3) times. So it's not a net win.

> Would probably also make sense to update frequent_items_with_threshold as a follow-up because here we still clone every qualifying item into owned Row's before the caller can filter

This may be possible. But now the returned rows are sorted so it seems not to be suitable to return an iterator.

If you mean we may change `Row<T>` to `Row<'a, T>`, as it's used by end users, I'm afraid that would increase usage complexity - we may review this when we have a concrete use case that proves its benefits.

tisonkun:
But `Row<'a, T>` is possible, anyway. I just need one more consideration and would prefer a real use case.

Something like:

```diff
index 49dce3b..6895e4a 100644
--- a/datasketches/src/frequencies/sketch.rs
+++ b/datasketches/src/frequencies/sketch.rs
@@ -57,17 +57,17 @@ pub enum ErrorType {
 ///
 /// Each row includes an estimate and upper and lower bounds on the true frequency.
 #[derive(Debug, Clone, PartialEq, Eq)]
-pub struct Row<T> {
-    item: T,
+pub struct Row<'a, T> {
+    item: &'a T,
     estimate: u64,
     upper_bound: u64,
     lower_bound: u64,
 }
 
-impl<T> Row<T> {
+impl<'a, T> Row<'a, T> {
     /// Returns the item value.
-    pub fn item(&self) -> &T {
-        &self.item
+    pub fn item(&self) -> &'a T {
+        self.item
     }
 
     /// Returns the estimated frequency.
@@ -378,10 +378,7 @@ impl<T: Eq + Hash> FrequentItemsSketch<T> {
     /// let rows = sketch.frequent_items(ErrorType::NoFalseNegatives);
     /// assert!(rows.iter().any(|row| *row.item() == 1));
     /// ```
-    pub fn frequent_items(&self, error_type: ErrorType) -> Vec<Row<T>>
-    where
-        T: Clone,
-    {
+    pub fn frequent_items(&self, error_type: ErrorType) -> impl Iterator<Item = Row<'_, T>> {
         self.frequent_items_with_threshold(error_type, self.offset)
     }
 
@@ -407,13 +404,9 @@ impl<T: Eq + Hash> FrequentItemsSketch<T> {
         &self,
         error_type: ErrorType,
         threshold: u64,
-    ) -> Vec<Row<T>>
-    where
-        T: Clone,
-    {
+    ) -> impl Iterator<Item = Row<'_, T>> {
         let threshold = threshold.max(self.offset);
-        let mut rows = vec![];
-        for (item, count) in self.hash_map.iter() {
+        self.hash_map.iter().filter_map(move |(item, count)| {
             let lower = count;
             let upper = count + self.offset;
             let include = match error_type {
@@ -421,16 +414,16 @@ impl<T: Eq + Hash> FrequentItemsSketch<T> {
                 ErrorType::NoFalsePositives => lower > threshold,
             };
             if include {
-                rows.push(Row {
-                    item: item.clone(),
+                Some(Row {
+                    item,
                     estimate: upper,
                     upper_bound: upper,
                     lower_bound: lower,
-                });
+                })
+            } else {
+                None
             }
-        }
-        rows.sort_by_key(|row| std::cmp::Reverse(row.estimate));
-        rows
+        })
     }
 
     fn maybe_resize_or_purge(&mut self) {
```

But this would drop the sorted character that requires users to sort on their own. I need to review the contract.

k3dom:
> This is proposed by my Codex at the first time. But I reject it because an active_entries iterator would do the activity check every time while we'd iterator it several (now 3) times. So it's not a net win.

Yeah agree, that is what I meant when I said it's probably not faster for large maps. For most map sizes the three scans should be cheap but sparse scans probably loose to collect-once.

> But Row<'a, T> is possible, anyway. I just need one more consideration and would prefer a real use case.

Agree, I don't have a use case for my project right now either, just thought I mention it because I noticed it while going over the API once again

ZENOTME (approved):

PR #145, opened 2026-07-06
feat(theta): add theta union

ZENOTME:
## Summary
- add stateful ThetaUnion and builder
- export ThetaUnion from the theta module
- add union coverage for exact, estimation, compact, seed mismatch, ordering, reset, and edge states

## Tests
- cargo test -p datasketches --features theta


datasketches/src/theta/union.rs
```diff
@@ -0,0 +1,233 @@
...
+        let nominal_num = 1usize << self.table.lg_nom_size();
+        if entries.len() > nominal_num {
+            let (_, kth, _) = entries.select_nth_unstable(nominal_num);
+            theta = *kth;
+            entries.truncate(nominal_num);
+        }
+
+        let ordered = ordered || (entries.len() == 1 && theta == MAX_THETA);
+
+        if ordered && entries.len() > 1 {
```
ariesdevil: seems no need check with `entries.len() > 1`, `sort_unstable()` no-op on single element array.

ariesdevil (approved):
Rest LGTM.

datasketches/src/theta/union.rs
```diff
@@ -0,0 +1,233 @@
...
+    ///
+    /// ```
+    /// # use datasketches::theta::ThetaUnion;
+    /// let _union = ThetaUnion::builder().lg_k(12).build();
+    /// ```
+    pub fn builder() -> ThetaUnionBuilder {
+        ThetaUnionBuilder::default()
+    }
+
+    /// Update this union with a given sketch.
```
tisonkun, line 56: Later we may review all seed_hash != seed_hash checks and add `# Errors` document.

tisonkun (approved):
Generally LGTM. Some follow-up suggestions inline, and:

I may prefer to name `result`/`result_with_order` to `to_sketch` as we do for the CpcSketch and HllSketch. Also, we may not need to have two methods but only one with a `ordered` param.

This is a breaking change since we already use the `result_with_order` API in `ThetaIntersection`, but it should be worthy to do the refactor.

tisonkun:
Now that #146 is merged. We may merge this as well.

Is there any logical conflict that should be resolved beforehand?

ZENOTME:
> Now that #146 is merged. We may merge this as well.
> 
> Is there any logical conflict that should be resolved beforehand?

Yes. Let me covert it as a generic version.

tisonkun (changes requested):
Thanks for your updates!

I'll go through all these changes and perhaps push a few commits to handle visibility issues.

Hopefully in days, but may be delay to next weekend.

datasketches/src/thetacommon/sketch_view.rs
```diff
@@ -0,0 +1,47 @@
...
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+use crate::theta_common::hash_table::RawHashTableEntry;
+
+pub(crate) mod private {
+    pub(crate) trait Sealed {}
```
ariesdevil: BTW: As @tisonkun mentioned on [X](https://x.com/tison1096/status/2071115508926345328?s=20), [`pub impl(crate) trait Foo {}`](https://rust-lang.github.io/rfcs/3323-restrictions.html) just landed in nightly version; maybe we can use this feature?

ariesdevil (approved):

datasketches/src/thetacommon/sketch_view.rs
tisonkun: This crate is a library and we should stay on stable toolchain. And we have an MSRV of 1.86.

ariesdevil: Make sense.

tisonkun:
Pushed a few commits to handle part of the structure and visibility issues. I'll handle for:

* `hash_table.rs`
* `sketch_view.rs`
* `union.rs`

... under `thetacommon` as well and call for a merge.

tisonkun (approved):

tisonkun:
There is one more pending issue: the two traits we made public in `thetacommon` are looking a bit strange right now. If you want to make them sealed or handle them differently, we probably need to refactor the tests in union.rs.

It feels very awkward to implement sealed traits for a struct inside the test module itself. We should take a step back and look at the overall scope of these two traits.

PR #146, opened 2026-07-09
refactor(theta): theta hash table into raw table, preparing for tuple reuse

ariesdevil:
## Summary

As @ZENOTME [mentioned](https://github.com/apache/datasketches-rust/pull/138#discussion_r3551839612), we should first extract a reusable hashtable trait for the upcoming tuple implementation.

related: #145

### DIFF
--- a/datasketches/src/theta/hash_table.rs
+++ b/datasketches/src/theta/hash_table.rs
@@ -16,20 +16,14 @@
 // under the License.
 
 use std::hash::Hash;
+use std::num::NonZeroU64;
 
-use crate::common::ResizeFactor;
-use crate::hash::MurmurHash3X64128;
-use crate::hash::compute_seed_hash;
-use crate::theta::HASH_TABLE_REBUILD_THRESHOLD;
-use crate::theta::HASH_TABLE_RESIZE_THRESHOLD;
-use crate::theta::MAX_THETA;
-use crate::theta::MIN_LG_K;
-
-/// Stride hash bits (7 bits for stride calculation)
-const STRIDE_HASH_BITS: u8 = 7;
-
-/// Stride mask
-const STRIDE_MASK: u64 = (1 << STRIDE_HASH_BITS) - 1;
+use super::raw_hash_table::RawHashTable;
+use super::raw_hash_table::RawHashTableEntry;
+#[cfg(test)]
+pub(crate) use super::raw_hash_table::starting_sub_multiple;
+#[cfg(test)]
+pub(crate) use super::raw_hash_table::starting_theta_from_sampling_probability;
 
 /// Specific hash table for theta sketch
 ///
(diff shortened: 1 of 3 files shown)


tisonkun (commented):
I remember we reject a general `Hashable` solution as in https://github.com/apache/datasketches-rust/pull/118.

But this PR anyway seems like to introduce `RawHashTableEntry` rather than `Hashable`.

I noticed that it's a crate-level trait. Would it benefits from being an `enum` rather than a `trait`?

ZENOTME:
> I remember we reject a general `Hashable` solution as in #118.
> 
> But this PR anyway seems like to introduce `RawHashTableEntry` rather than `Hashable`.
> 
> I noticed that it's a crate-level trait. Would it benefits from being an `enum` rather than a `trait`?

enum means that entry can be different type in single hash table. But in our case, hash table is reuse as 
```
type ThetaHashTable = RawHashTable<ThateHashEntry>

type TupleHashTable = RawHashTable<TupleHashEntry>
```

Seems trait is more clear.


> I remember we reject a general Hashable solution as in https://github.com/apache/datasketches-rust/pull/118.

This hash table is for reuse logic between tuple and theta sketch. Not export to user.

tisonkun:
SGTM.

We can merge #145 now and rebase this PR onto it and continue the review :D

ZENOTME:
> SGTM.
> 
> We can merge #145 now and rebase this PR onto it and continue the review :D

We can merge this first actually. Otherwise, this PR need to rewrite union. 🤣

datasketches/src/theta/hash_table.rs
```diff
@@ -38,130 +32,26 @@ const STRIDE_MASK: u64 = (1 << STRIDE_HASH_BITS) - 1;
...
+}
 
-    /// Hash a value with the table seed and return the hash.
-    fn hash<T: Hash>(&self, value: T) -> u64 {
-        let mut hasher = MurmurHash3X64128::with_seed(self.hash_seed);
-        value.hash(&mut hasher);
-        let (h1, _) = hasher.finish128();
-        h1 >> 1 // To make it compatible with Java version
+impl ThetaEntry {
+    fn new(hash: u64) -> Option<Self> {
```
ZENOTME: How about `fn new(hash: u64) -> Self`. Seem it only use internally and we gurantee that it's always valid. 🤔 Panic when hash is not zero.

ariesdevil: Nice catch, fixed.

datasketches/src/theta/mod.rs
```diff
@@ -52,14 +53,18 @@ pub use self::sketch::ThetaSketchBuilder;
...
 /// Resize threshold (0.5 = 50% load factor)
-const HASH_TABLE_RESIZE_THRESHOLD: f64 = 0.5;
+pub(crate) const HASH_TABLE_RESIZE_THRESHOLD: f64 = 0.5;
 /// Rebuild threshold (15/16 = 93.75% load factor)
-const HASH_TABLE_REBUILD_THRESHOLD: f64 = 15.0 / 16.0;
+pub(crate) const HASH_TABLE_REBUILD_THRESHOLD: f64 = 15.0 / 16.0;
+/// Stride hash bits (7 bits for stride calculation)
+pub(crate) const STRIDE_HASH_BITS: u8 = 7;
+/// Stride mask
+pub(crate) const STRIDE_MASK: u64 = (1 << STRIDE_HASH_BITS) - 1;
```
tisonkun, line 70: These `pub(crate)` should be useless. They are limited to the `theta` module and can be already accessed by any sub module.

datasketches/src/theta/raw_hash_table.rs
```diff
@@ -0,0 +1,400 @@
...
+    // This can be false even when `num_retained` is 0.
+    pub(crate) is_empty: bool,
+
+    pub(crate) theta: u64,
+
+    pub(crate) entries: Vec<Option<E>>,
+
+    // Number of retained non-zero hashes currently stored in `entries`.
+    pub(crate) num_retained: usize,
+}
```
tisonkun, line 60: These `pub(crate)` can be `pub(super)` as they are limited to the `theta` module, never accesed by other modules.

datasketches/src/theta/hash_table.rs
```diff
@@ -16,20 +16,14 @@
...
-const STRIDE_HASH_BITS: u8 = 7;
-
-/// Stride mask
-const STRIDE_MASK: u64 = (1 << STRIDE_HASH_BITS) - 1;
+use super::raw_hash_table::RawHashTable;
+use super::raw_hash_table::RawHashTableEntry;
+#[cfg(test)]
+pub(crate) use super::raw_hash_table::starting_sub_multiple;
+#[cfg(test)]
+pub(crate) use super::raw_hash_table::starting_theta_from_sampling_probability;
```
tisonkun: nit: may prefer to `use crate:xxx` since it's more refactor-friendly. But fair enough to gradually converge as we don't have a linter rule to guard it now.

datasketches/src/theta/mod.rs
ariesdevil: These are defined as `pub(crate)` because they will be used in subsequent PRs in the stacked PRs; the tuple implementation requires using these constants. 
If you feel this PR shouldn't be written this way, I can make the changes.

FYI: https://github.com/apache/datasketches-rust/pull/138/changes#diff-f3568728db442efe8852306fd065409eb9c4286921170ef66c3cde0b9d452be4R31-R33

datasketches/src/theta/raw_hash_table.rs
ariesdevil: ditto

ZENOTME (approved):
LGTM

datasketches/src/theta/mod.rs
tisonkun: Fair enough. Then let's keep the `pub(crate)` structure now and do a review (refactor) later when all the function code landed.

tisonkun: I guess we can finally move the shared code to a `datasketches/thetacommon` module later. This is how datasketches-java organizes code also.

But let's review this then.

tisonkun (approved):
LGTM. Further improvement suggestions can be addressed once we get all the functions landed.

PR #147, opened 2026-07-10
feat: add tuple theta implementation and tests based on raw hashtable

ariesdevil:
Wait for #145 

## Summary

## Note

### DIFF
--- a/datasketches/Cargo.toml
+++ b/datasketches/Cargo.toml
@@ -45,6 +45,7 @@ frequencies = []
 hll = []
 tdigest = []
 theta = []
+tuple = []
 
 [dev-dependencies]
 googletest = { workspace = true }
--- a/datasketches/src/codec/decode.rs
+++ b/datasketches/src/codec/decode.rs
@@ -38,6 +38,16 @@ impl SketchSlice<'_> {
         self.slice.set_position(pos + n);
     }
 
+    /// Returns the not-yet-read portion of the underlying slice.
+    ///
+    /// Useful for handing the remaining bytes to a variable-length decoder that reports how many
+    /// bytes it consumed; pair it with [`advance`](Self::advance).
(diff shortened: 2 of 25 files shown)


datasketches/src/theta/mod.rs
```diff
@@ -46,7 +46,17 @@ mod raw_hash_table;
...
+// raw table directly.
+#[cfg(all(feature = "tuple", test))]
+pub(crate) use self::hash_table::starting_sub_multiple;
+#[cfg(all(feature = "tuple", test))]
+pub(crate) use self::hash_table::starting_theta_from_sampling_probability;
 pub use self::intersection::ThetaIntersection;
+#[cfg(feature = "tuple")]
+pub(crate) use self::raw_hash_table::RawHashTable;
+#[cfg(feature = "tuple")]
+pub(crate) use self::raw_hash_table::RawHashTableEntry;
```
tisonkun: #145 moves these symbols to `thetacommon`, we'd avoid these `pub(crate) use` now.

datasketches/Cargo.toml
```diff
@@ -45,6 +45,7 @@ frequencies = []
 hll = []
 tdigest = []
 theta = []
+tuple = ["theta"]
```
tisonkun: We can enable `thetacommon` on either `theta` or `tuple`, and thus need not this dep.

```rust
#[cfg(any(feature = "theta", feature = "tuple"))]
pub mod thetacommon;
```

ZENOTME:
Thanks @ariesdevil ! This PR is a littble big to review actually, could you help to split it into independent part to make it easier to review:
1. Tuple sketch
2. Union operation 
3. Intersection operation 
4. a not b operation

ariesdevil:
@ZENOTME Done #156

PR #151, opened 2026-07-19
refactor: encapsulate read-only Theta sketch views

ZENOTME:
## Summary

- keep RawThetaSketchView crate-private while preserving the raw algorithm abstraction
- add a transparent ThetaSketchView<V> that can only be created by supported sketches
- expose the six read-only view methods and accept views in Theta union/intersection
- add reference forwarding plus From conversions for mutable and compact Theta sketches

## Validation

- cargo test -p datasketches --features theta
- cargo clippy --workspace --all-features --all-targets -- -D warnings
- RUSTDOCFLAGS="-D warnings" cargo doc -p datasketches --features theta --no-deps

### DIFF
--- a/datasketches/src/theta/intersection.rs
+++ b/datasketches/src/theta/intersection.rs
@@ -20,7 +20,9 @@ use crate::error::Error;
 use crate::hash::DEFAULT_UPDATE_SEED;
 use crate::theta::CompactThetaSketch;
 use crate::theta::ThetaSketchView;
+use crate::theta::hash_table::ThetaEntry;
 use crate::theta::hash_table::ThetaHashTable;
+use crate::thetacommon::RawThetaSketchView;
 use crate::thetacommon::constants::HASH_TABLE_REBUILD_THRESHOLD;
 use crate::thetacommon::constants::MAX_THETA;
 
@@ -61,7 +63,12 @@ impl ThetaIntersection {
     /// The intersection can be viewed as starting from the "universe" set,
     /// and every update can reduce the current set to leave the overlapping
     /// subset only.
-    pub fn update<S: ThetaSketchView>(&mut self, sketch: &S) -> Result<(), Error> {
+    #[allow(private_bounds)]
+    pub fn update<V>(&mut self, sketch: ThetaSketchView<V>) -> Result<(), Error>
+    where
+        V: RawThetaSketchView<ThetaEntry>,
+    {
+        let sketch = sketch.inner;
         let new_default_table = |table: &ThetaHashTable| {
             ThetaHashTable::from_raw_parts(
                 0,
--- a/datasketches/src/theta/sketch.rs
+++ b/datasketches/src/theta/sketch.rs
@@ -52,11 +52,47 @@ use crate::thetacommon::constants::MIN_LG_K;
 
 /// Read-only view for Theta sketches.
 ///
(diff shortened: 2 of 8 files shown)


datasketches/src/theta/intersection.rs
```diff
@@ -61,7 +63,12 @@ impl ThetaIntersection {
     /// The intersection can be viewed as starting from the "universe" set,
     /// and every update can reduce the current set to leave the overlapping
     /// subset only.
-    pub fn update<S: ThetaSketchView>(&mut self, sketch: &S) -> Result<(), Error> {
+    #[allow(private_bounds)]
+    pub fn update<V>(&mut self, sketch: ThetaSketchView<V>) -> Result<(), Error>
+    where
+        V: RawThetaSketchView<ThetaEntry>,
+    {
```
tisonkun, line 70: This would make the document of `update` cannot linked to `RawThetaSketchView` and remain a dead link on docs.rs.

Let me think of how to expose the symbols here.

tisonkun (changes requested):
Looks like an incorrect abstraction.

I think we can review and merge the tuple sketches code and later try to factor out the abstraction.

So far, I don't see a concrete reason for all these abstratcions. We can inline the `ThetaSketchView` if this is the only impl.

datasketches/src/theta/intersection.rs
ZENOTME: To avoid trait here, we can use enum to dispatch.🤔

ZENOTME: Another way is to export using new trait like `TupleSketchView`. We don't need to test it, and it can be sealed::Sealed. And thetacommon can internal and not export to user. 

```
pub trait TupleSketchView<S>: sealed::Sealed {
    /// Return the 16-bit seed hash.
    fn seed_hash(&self) -> u16;

    /// Return theta as a `u64` threshold.
    fn theta(&self) -> u64;

    /// Return whether this sketch has not received any updates.
    fn is_empty(&self) -> bool;

    /// Return whether retained entries are ordered by ascending hash.
    fn is_ordered(&self) -> bool;

    /// Return an iterator over retained entries.
    fn iter(&self) -> impl Iterator<Item = TupleEntry<S>> + '_;

    /// Return the number of retained entries.
    fn num_retained(&self) -> usize;
}

impl<S, T> RawThetaSketchView<TupleEntry<S>> for T
where
    T: TupleSketchView<S> + ?Sized,
{
    fn seed_hash(&self) -> u16 {
        TupleSketchView::seed_hash(self)
    }

    fn theta(&self) -> u64 {
        TupleSketchView::theta(self)
    }

    fn is_empty(&self) -> bool {
        TupleSketchView::is_empty(self)
    }

    fn is_ordered(&self) -> bool {
        TupleSketchView::is_ordered(self)
    }

    fn iter(&self) -> impl Iterator<Item = TupleEntry<S>> + '_ {
        TupleSketchView::iter(self)
    }

    fn num_retained(&self) -> usize {
        TupleSketchView::num_retained(self)
    }
}
```

tisonkun:
@ariesdevil You refactor the related interfaces in this PR in #166 .

I suggest you can discuss with @ZENOTME to see whether or not these `thetacommon` traits should be sealed and how to settle down a proper design.

PR #152, opened 2026-07-19
feat: add tuple sketch implementation

ariesdevil:
## Summary

## Note

### DIFF
--- a/datasketches/src/lib.rs
+++ b/datasketches/src/lib.rs
@@ -45,8 +45,10 @@ pub mod hll;
 pub mod tdigest;
 #[cfg(feature = "theta")]
 pub mod theta;
-#[cfg(feature = "theta")]
+#[cfg(any(feature = "theta", feature = "tuple"))]
 pub mod thetacommon;
+#[cfg(feature = "tuple")]
+pub mod tuple;
 
 // common modules
 pub mod codec;
--- a/datasketches/src/lib.rs
+++ b/datasketches/src/lib.rs
@@ -46,6 +46,7 @@ pub mod tdigest;
 #[cfg(feature = "theta")]
 pub mod theta;
 #[cfg(any(feature = "theta", feature = "tuple"))]
+#[allow(dead_code)] // some utilities are only used for certain sketches
 pub mod thetacommon;
 #[cfg(feature = "tuple")]
 pub mod tuple;
--- a/datasketches/src/tuple/hash_table.rs
+++ b/datasketches/src/tuple/hash_table.rs
@@ -22,12 +22,11 @@ use crate::thetacommon::RawHashTableEntry;
 use crate::thetacommon::hash_table::RawHashTable;
 
 /// A retained entry in a Tuple sketch: a hash key together with its associated summary.
-///
-/// The hash is stored as [`NonZeroU64`] (hash 0 is screened out before insertion), so
-/// `Option<TupleEntry<S>>` keeps the niche and takes no more space than `TupleEntry<S>` itself —
-/// the same layout the Theta table gets from its `NonZeroU64` entry.
 #[derive(Debug, Clone)]
 pub struct TupleEntry<S> {
+    // Note that this field is stored as `NonZeroU64` (hash 0 is screened out before insertion),
+    // so `Option<TupleEntry<S>>` keeps the niche and takes no more space than `TupleEntry<S>`
+    // itself.
     hash: NonZeroU64,
     summary: S,
 }
@@ -99,7 +98,7 @@ impl<S> TupleHashTable<S> {
         })
     }
 
-    /// Get iterator over retained entries as `(hash, &summary)` pairs.
+    /// Returns an iterator over retained entries as `(hash, &summary)` pairs.
     pub fn iter(&self) -> impl Iterator<Item = (u64, &S)> + '_ {
         self.iter_entries()
             .map(|entry| (entry.hash.get(), &entry.summary))
--- a/datasketches/src/tuple/hash_table.rs
+++ b/datasketches/src/tuple/hash_table.rs
@@ -116,21 +116,23 @@ mod tests {
     use crate::thetacommon::hash_table::starting_sub_multiple;
     use crate::thetacommon::hash_table::starting_theta_from_sampling_probability;
 
-    /// Inserts a key with count-style summary semantics: a new key starts at 1, a repeated key
-    /// increments the retained count. Returns true if a new entry was created.
-    fn insert(table: &mut TupleHashTable<u64>, value: impl Hash) -> bool {
-        table.try_insert(value, |existing| match existing {
-            Some(count) => {
-                *count += 1;
-                None
-            }
-            None => Some(1),
-        })
-    }
+    impl TupleHashTable<u64> {
+        /// Inserts a key with count-style summary semantics: a new key starts at 1, a repeated key
+        /// increments the retained count. Returns true if a new entry was created.
+        fn insert(&mut self, value: impl Hash) -> bool {
+            self.try_insert(value, |existing| match existing {
+                Some(count) => {
+                    *count += 1;
+                    None
+                }
+                None => Some(1),
+            })
+        }
 
-    /// Collect retained `(hash, count)` pairs.
-    fn collect(table: &TupleHashTable<u64>) -> Vec<(u64, u64)> {
-        table.iter().map(|(hash, &count)| (hash, count)).collect()
+        /// Returns the retained `(hash, count)` pairs.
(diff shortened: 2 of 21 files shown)


datasketches/src/tuple/sketch.rs
```diff
@@ -0,0 +1,1020 @@
...
+/// ```
+/// # use datasketches::tuple::UpdatableTupleSketch;
+/// let mut sketch = UpdatableTupleSketch::<u64>::builder().build();
+/// sketch.update("apple", 1);
+/// sketch.update("apple", 1);
+/// assert!(sketch.estimate() >= 1.0);
+/// assert_eq!(sketch.num_retained(), 1);
+/// ```
+#[derive(Debug)]
+pub struct UpdatableTupleSketch<S, P = DefaultUpdatePolicy> {
```
ZENOTME: UpdatableTupleSketch -> TupleSketch

datasketches/src/tuple/sketch.rs
```diff
@@ -0,0 +1,1020 @@
...
+    }
+}
+
+/// Compact (immutable) Tuple sketch.
+///
+/// This is the serialization-friendly form: a compact array of `(hash, summary)` entries plus theta
+/// and a 16-bit seed hash. It can be ordered (sorted ascending by hash) or unordered.
+#[derive(Clone, Debug)]
+pub struct CompactTupleSketch<S> {
+    entries: Vec<(u64, S)>,
```
ZENOTME: Why not use `Vec<TupleEntry<S>>` direcly.

datasketches/src/tuple/hash_table.rs
```diff
@@ -0,0 +1,480 @@
...
+        self.hash.get()
+    }
+
+    /// Returns the summary stored in this entry.
+    pub fn summary(&self) -> &S {
+        &self.summary
+    }
+
+    /// Splits this entry into its `(hash, summary)` parts.
+    pub(super) fn into_parts(self) -> (u64, S) {
```
ZENOTME: Maybe we don't need it if CompactTuple store TupleEntry.

datasketches/src/tuple/hash_table.rs
```diff
@@ -0,0 +1,480 @@
...
+/// This is the Theta sketch hash table extended so that each retained key carries a user-defined
+/// summary. It maintains an array with capacity up to 2^lg_max_size:
+/// * Before it reaches the max capacity, it will extend the array based on resize_factor.
+/// * After it reaches the capacity bigger than 2^lg_nom_size, every time the number of entries
+///   exceeds the threshold, it will rebuild the table: only keep the min 2^lg_nom_size entries and
+///   update the theta to the k-th smallest entry.
+///
+/// Unlike the Theta hash table, when a key is inserted that already exists, the incoming update is
+/// merged into the existing summary rather than discarded.
+pub(super) type TupleHashTable<S> = RawHashTable<TupleEntry<S>>;
```
ZENOTME, line 61: > It maintains an array with capacity up to 2^lg_max_size:
/// * Before it reaches the max capacity, it will extend the array based on resize_factor.
/// * After it reaches the capacity bigger than 2^lg_nom_size, every time the number of entries
///   exceeds the threshold, it will rebuild the table: only keep the min 2^lg_nom_size entries and
///   update the theta to the k-th smallest entry.

I think this comment may better in RawHashTable rather a specific comment in RawHashTable. We can dicard it for now in this place.

datasketches/src/tuple/hash_table.rs
```diff
@@ -0,0 +1,480 @@
...
+        self.hash.get()
+    }
+}
+
+impl<S> TupleHashTable<S> {
+    /// Hashes a key and inserts or updates its summary via a single callback.
+    ///
+    /// See [`upsert`](Self::upsert) for the callback contract. Returns true if a new entry was
+    /// created, false if the key already existed or the hash was screened out by theta.
+    pub fn update<T, F>(&mut self, key: T, f: F) -> bool
```
ZENOTME: How about named it in consistent way as theta:
1. try_insert 
2. try_insert_hash

datasketches/src/tuple/hash_table.rs
```diff
@@ -0,0 +1,480 @@
...
+        T: Hash,
+        F: FnOnce(Option<&mut S>) -> Option<S>,
+    {
+        let hash = self.hash(key);
+        self.upsert(hash, f)
+    }
+
+    /// Inserts or updates the summary slot for a pre-hashed key.
+    ///
+    /// The callback `f` is invoked with the current summary for `hash`:
```
ZENOTME: This contract comment here looks more appropriate in `RawHashTbale::upsert_entry`

datasketches/src/tuple/policy.rs
```diff
@@ -0,0 +1,108 @@
...
+// under the License.
+
+//! Policies describing how summaries are created, updated, and combined.
+//!
+//! A Tuple sketch keeps a user-defined summary `S` next to every retained key. The behavior of a
+//! summary is supplied externally through policy objects rather than baked into the summary type
+//! itself, so the same summary type (for example a plain `u64` or a `Vec<f64>`) can be driven by
+//! different behaviors and can carry per-instance configuration (such as the number of values in an
+//! array-of-doubles summary).
+//!
```
ZENOTME: > //!
//! This mirrors the policy approach used by the C++ implementation
//! (`default_tuple_update_policy`, `default_tuple_union_policy`) and the Java
//! `SummaryFactory` / `SummarySetOperations` interfaces.

Not sure whether it's need.

datasketches/src/tuple/policy.rs
```diff
@@ -0,0 +1,108 @@
...
+
+use std::ops::AddAssign;
+
+/// Defines how a summary is created and how update values are folded into it.
+///
+/// This is used by the update tuple sketch. `S` is the stored summary type and `U` is the type of
+/// the update value, which may be a borrowed type such as `&[f64]`.
+///
+/// Corresponds to C++ `default_tuple_update_policy<Summary, Update>` and Java
+/// `UpdatableSummary<U>` together with `SummaryFactory`.
```
ZENOTME: > /// Corresponds to C++ `default_tuple_update_policy<Summary, Update>` and Java
/// `UpdatableSummary<U>` together with `SummaryFactory`.

Not sure whether it's need. It's easy to be inconsistent across different language.

ariesdevil:
@ZENOTME Fixed, PTAL again.

datasketches/src/tuple/serde.rs
```diff
@@ -0,0 +1,164 @@
...
+                        buf.copy_from_slice(&bytes[..std::mem::size_of::<$t>()]);
+                        <$t>::from_le_bytes(buf)
+                    }
+                }
+            )*
+        };
+    }
+
+    impl_le_bytes!(u32, u64, i32, i64, f32, f64);
+}
```
tisonkun: Why we need this? The `SketchSlice` and `SketchBytes` in the `codec` module should implement all these functionalities. Or we should add new utilities there instead of randomly add some code in dedicated skecthes modules.

tisonkun (commented):
Generally LGTM. Comments inline.

datasketches/src/tuple/serde.rs
ariesdevil: `SketchBytes` / `SketchSlice` only offer per-type concrete methods (`write_u64_le`, `read_f32_le`, ...), but `PrimitiveSummarySerde` is a generic `impl<S> SummarySerde<S>` — it needs a trait to dispatch on the type's LE width and encoding, so the abstraction itself can't go away. 
Moved `LeBytes` to `codec/primitive.rs` as a `pub(crate)` utility instead of hiding it in the tuple module; frequencies' `impl_primitive!` could migrate to it later if desired.

tisonkun (approved):
LGTM.

I pushed several commits to do post fine tune.

I'm going to submit a new PR to rework the `Builder` interface. Basically, no need for several construction point and avoid some strange interface change.

datasketches/src/lib.rs
```diff
@@ -46,6 +46,7 @@ pub mod tdigest;
 #[cfg(feature = "theta")]
 pub mod theta;
 #[cfg(any(feature = "theta", feature = "tuple"))]
+#[allow(dead_code)] // some utilities are only used for certain sketches
```
tisonkun, line 49: To be reviewed once the full tuple sketch implementation landed.

ariesdevil:
Thx for your patient review, let's move on.

tisonkun:
> I'm going to submit a new PR to rework the `Builder` interface. Basically, no need for several construction point and avoid some strange interface change.

See https://github.com/apache/datasketches-rust/pull/158

PR #154, opened 2026-07-19
feat: support tuple intersection

ariesdevil:
based on #152 and #153

PR #153, opened 2026-07-19
feat: support tuple union

ariesdevil:
based on #152

PR #155, opened 2026-07-19
feat: support theta and tuple a-not-b

ariesdevil:
based on #152 #153 #154

### DIFF
--- a/datasketches/src/theta/a_not_b.rs
+++ b/datasketches/src/theta/a_not_b.rs
@@ -1,0 +1,95 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+//! Theta sketch set difference (`A and not B`).
+//!
+//! [`ThetaAnotB`] computes the set difference of two Theta sketches: the keys retained in `A`
+//! that are not present in `B`. The logic lives in the shared raw operator (`RawAnotB`) that also
+//! drives the Tuple a-not-B; Theta entries carry no summary, so nothing needs to be combined.
+
+use crate::error::Error;
+use crate::hash::DEFAULT_UPDATE_SEED;
+use crate::theta::CompactThetaSketch;
+use crate::theta::ThetaSketchView;
+use crate::thetacommon::a_not_b::RawAnotB;
+
+/// Set difference operator (`A and not B`) for Theta sketches.
+///
+/// This is a stateless operator (other than the seed): each call to [`compute`](Self::compute)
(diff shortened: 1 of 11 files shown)


tisonkun:
#154 is merged now.

I wonder if we'd have a `tuple_a_not_b_test.rs` integration test. And some common test code may be de-duplicate later.

datasketches/src/tuple/a_not_b.rs
```diff
@@ -0,0 +1,472 @@
...
+///
+/// let mut b = TupleSketchBuilder::new(update_policy).build();
+/// b.update("banana", 1);
+///
+/// let a_not_b = TupleAnotB::new_with_default_seed();
+/// let result = a_not_b.compute(&a, &b, true).unwrap();
+/// assert_eq!(result.num_retained(), 1); // only "apple" survives
+/// ```
+#[derive(Debug, Clone, Copy)]
+pub struct TupleAnotB {
```
ZENOTME, line 54: Should we have shared `RawAnotB` in thetacommon?

datasketches/tests/common.rs
tisonkun, line 1: I have some ideas on factoring out this test utility support module. Will do once this PR gets merged.

tisonkun: ... to avoid too much `#[cfg(feature = "...")]` and `#[allow(dead_code)]`.

datasketches/src/theta/a_not_b.rs
tisonkun, line 1: `a_not_b` comes from datasketches-cpp and datasketches-java?

@leerho @proost what if we name this set operation as `difference` or `exception`, since `a_not_b` is somewhat strange to me. Or is it actually an idiom in this domain?

tisonkun (commented):
Generally LGTM.

Anything keeps this PR a draft before it's ready for review and merge?

tisonkun (approved):
Minor comments can be handled later.

datasketches/src/theta/a_not_b.rs
tisonkun: Based on the result, it looks like the same as set difference.

@ariesdevil What do you think?

leerho: **a_not_b** **is** a set difference.  However the name **a_not_b** is a mnemonic that tells you the direction of the difference.  

In the math of boolean logic one would write 

A \ B := A AND NOT B = **$A \cup !B$** 
or
B \ A := B AND NOT A = **$B \cup !A$**

Where "\\" is the difference operator.  
Note that A \ B != B \ A !

Although C++ allows user defined operators most languages do not. To call the difference function, we need to indicate the direction. We chose to do it by the position of the arguments, and the name of the function is a mnemonic of the direction.

Sketch result a_not_b(Sketch A, Sketch B);

tisonkun: @leerho Thanks for your information!

Yes the Rust impl also defines the API as:

```rust
let update_policy = DefaultUpdatePolicy::<u64>::default();
let mut a = TupleSketchBuilder::new(update_policy).build();
a.update("apple", 1);
a.update("banana", 1);

let mut b = TupleSketchBuilder::new(update_policy).build();
b.update("banana", 1);

let a_not_b = TupleAnotB::default();
let result = a_not_b.compute(&a, &b, true).unwrap();
```

So I generally agree that `a_not_b` provides more information and avoid the confusion of op direction.

Even if we can add methods or operator overrides to `ThetaSketch`/`TupleSketch`, it is not symmetry to the union and intersection interface.

One minor comment: any typical reason to name it `TupleAnotB` rather than `TupleANotB`? Weird `CamelCase` at first glance.

leerho: @tisonkun,
You are correct, a_not_b is not symmetric, because the math operation it emulates is not symmetric.  Yes, and the CamelCase is not perfect.  But other languages are free to spell the function differently if they like.

PR #160, opened 2026-07-24
ci: download snapshots from apache/dataSketches-tck

### LINKED ISSUE #54: Cache generate_serialization_test_data.py's output to speed up CI
To speed up the CI times we should [cache the sketch files in CI](https://docs.github.com/en/actions/reference/workflows-and-actions/dependency-caching), keying them by commit hash of the repositories that generated them.

If datasketches-cpp or datasketches-java weren't updated in between CI runs there is no value in re-generating the serialised files so we should leverage the CI caching logic to make it more efficient.

Phoenix500526:
This avoids rebuilding Java and C++ in every CI job.

CLOSES #54

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -91,11 +91,6 @@ jobs:
       - uses: Swatinem/rust-cache@v2
         with:
           cache-bin: "false"
-      - name: Setup Java
-        uses: actions/setup-java@v5
-        with:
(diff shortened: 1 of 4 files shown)


tisonkun (approved):
Thanks for your conrtibution @Phoenix500526 !

Generally LGTM. The script name can be changed to fit its real logic now and I may do some code tidy for the script post merge.

PR #163, opened 2026-07-25
chore: simplify test support code

tisonkun:
This is for https://github.com/apache/datasketches-rust/pull/155#discussion_r3649620143.

cc @ariesdevil

PR #164, opened 2026-07-25
refactor: align seed and internal naming

tisonkun:
## Summary

This PR applies the approved subset of the naming-drift audit. It establishes an explicit distinction between a full 64-bit `seed` and its derived 16-bit `seed_hash`, standardizes the set-operation constructor style, and corrects several misleading or non-idiomatic internal names.

## Breaking constructor changes

Theta and Tuple set-operation constructors now follow the same convention as the other sketches:

- `ThetaIntersection::default()` uses the default seed; `ThetaIntersection::with_seed(seed)` accepts a custom seed.
- `ThetaAnotB::default()` uses the default seed; `ThetaAnotB::with_seed(seed)` accepts a custom seed.
- `TupleIntersection::new(policy)` uses the default seed; `TupleIntersection::with_seed(policy, seed)` accepts a custom seed.
- `TupleAnotB::default()` uses the default seed; `TupleAnotB::with_seed(seed)` accepts a custom seed.

The previous `new(seed)` and `new_with_default_seed` forms are removed and the migration is recorded in `CHANGELOG.md`.

## Commit-by-commit rationale

1. `7f15f37` — **S1: define seed terminology.** Documents that `seed` is the full 64-bit hashing value, `seed_hash` is only the derived 16-bit compatibility fingerprint, and `hash_seed` must not be used as a synonym.
2. `643a358` — **S2: rename the full RawHashTable seed.** Changes the internal `hash_seed` field, parameters, and accessor to `seed`, removing the only third synonym while leaving Count-Min row `hash_seeds` intact.
3. `0646322` — **S3: standardize seed-hash errors.** Converts the remaining CPC and A-not-B outliers to the common `incompatible seed hash` wording while retaining A/B input context.
4. `1278d7b` — **C1: standardize set-operation constructors.** Introduces consistent default-seed and custom-seed entry points across Theta and Tuple intersection and A-not-B operators; updates examples, tests, and the changelog.
5. `d7db047` — **P1: align CPC confidence parameter names.** Uses `num_std_dev` on public CPC sketch and wrapper methods while retaining the mathematical `kappa` name inside estimator internals.
6. `d073774` — **P3: align ThetaUnionBuilder parameter names.** Renames local parameters to `factor` and `probability`, matching the equivalent Theta and Tuple builders without changing method names or behavior.
7. `9611f37` — **P4: clarify the raw A-not-B type.** Renames the shared internal operator from `RawAnotB` to `RawThetaAnotB`, making its relationship to `RawThetaUnion` and `RawThetaIntersection` explicit.
8. `2a01ef3` — **I2: name the hash-table threshold accurately.** Replaces `get_capacity` with `capacity_threshold`, since the value controls resize or rebuild rather than reporting allocated vector capacity.
9. `01914e5` — **I3: use the approved entry lookup name.** Replaces `get_entry` with `entry`, following the selected Rust-style accessor name.
10. `d45955c` — **I4: expose intersection state directly.** Renames the internal `is_valid` field to `has_result`, matching the actual state-machine meaning and the existing public method.
11. `b0d38fb` — **I5: use the plural byte-count name.** Renames `byte_used` to `bytes_used` and corrects its documentation because the method returns a count.
12. `b8d7242` — **I6: follow mutable-accessor convention.** Renames `mut_surprising_value_table` to `surprising_value_table_mut`, matching standard Rust `_mut` suffix usage.
13. `76a78b9` — **I7: correct the raw-view documentation.** Describes the view as shared by Theta-family union, intersection, and A-not-B operations rather than union alone.
14. `1264c3a` — **Use Default for zero-configuration operators.** Replaces zero-argument `new()` methods with `Default` implementations on `ThetaIntersection`, `ThetaAnotB`, and `TupleAnotB`, matching the existing CPC, T-Digest, and builder conventions. `TupleIntersection::new(policy)` remains because its policy is required.

## Validation

- [x] `cargo x test`
- [x] `cargo x lint`
- [x] Targeted Theta/Tuple intersection and A-not-B tests
- [x] Rebased onto `main` after #163

## Explicitly deferred

This PR does not change the `RawThetaSketchView::theta` or `iter` API, does not remove cached `seed_hash` fields, does not rename `Coupon::from_hash`, and does not include any other items from the audit that were not explicitly approved.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ All significant changes to this project will be documented in this file.
 ### Breaking changes
 
 * Remove `ThetaSketch::builder`, `ThetaUnion::builder`, and `TupleSketch::builder`. Construct `ThetaSketchBuilder`, `ThetaUnionBuilder`, and `TupleSketchBuilder` with `Default::default` instead.
+* Standardize Theta and Tuple set-operation constructors so `new` uses the default seed and
+  `with_seed` accepts a custom seed. This replaces the previous `new(seed)` and
+  `new_with_default_seed` methods on intersection and A-not-B operators.
 
 ### New features
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,9 +7,10 @@ All significant changes to this project will be documented in this file.
 ### Breaking changes
 
 * Remove `ThetaSketch::builder`, `ThetaUnion::builder`, and `TupleSketch::builder`. Construct `ThetaSketchBuilder`, `ThetaUnionBuilder`, and `TupleSketchBuilder` with `Default::default` instead.
-* Standardize Theta and Tuple set-operation constructors so `new` uses the default seed and
-  `with_seed` accepts a custom seed. This replaces the previous `new(seed)` and
-  `new_with_default_seed` methods on intersection and A-not-B operators.
+* Standardize Theta and Tuple set-operation constructors. Zero-configuration operators use
+  `Default::default()`, `TupleIntersection::new(policy)` uses the default seed, and `with_seed`
+  accepts a custom seed. This replaces the previous `new(seed)` and `new_with_default_seed`
+  methods.
 
 ### New features
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -83,6 +83,12 @@ Install the extra tools with:
 cargo install taplo-cli typos-cli hawkeye
 ```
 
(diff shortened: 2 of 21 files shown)


tisonkun:
cc @ariesdevil in case you're working on theta/tuple sketches as well. This can be a code/logical conflict.

ZENOTME (approved):

ariesdevil:
Thx for informing me:)

PR #165, opened 2026-07-25
refactor: align theta view and coupon naming

tisonkun:
## Summary

This PR applies the two approved naming changes from the follow-up audit:

- rename `RawThetaSketchView::theta()` to `theta64()`
- rename `Coupon::from_hash()` directly to `from_value()` (option 8B)

The cached `seed_hash` fields remain unchanged following the final decision on item 6. The raw-view visibility strategy (item 1) and changelog work (item 5) remain deferred.

## Commit-by-commit rationale

1. `2aabb8a` — **Align the raw Theta threshold name.** Renames `RawThetaSketchView::theta()` to `theta64()` and updates every implementation and set-operation call site. The suffix makes the raw 64-bit threshold representation explicit and avoids conflating it with higher-level theta terminology. This is a naming-only change; threshold values and algorithms are unchanged.
2. `5baab5e` — **Name the Coupon constructor after its input.** Renames `Coupon::from_hash(value)` to `Coupon::from_value(value)` across the public API, documentation, examples, internal callers, and tests. The method hashes its generic input internally, so `from_hash` incorrectly implied that callers supplied an already-computed hash. Per option 8B, this is a direct rename without a deprecated compatibility alias.

## Compatibility notes

Both commits rename public methods. The changelog entry is intentionally deferred until the release review requested in item 5.

## Validation

- [x] `cargo x test`
- [x] `cargo x lint`
- [x] `git diff --check origin/main...HEAD`

### DIFF
--- a/datasketches/src/hll/mod.rs
+++ b/datasketches/src/hll/mod.rs
@@ -175,7 +175,7 @@ const RESIZE_DENOMINATOR: u32 = 4;
 ///
 /// ```
 /// # use datasketches::hll::{HllSketch, HllType, Coupon};
-/// let c = Coupon::from_hash("hello");
+/// let c = Coupon::from_value("hello");
 ///
 /// let mut sketch1 = HllSketch::new(10, HllType::Hll8);
 /// let mut sketch2 = HllSketch::new(12, HllType::Hll8);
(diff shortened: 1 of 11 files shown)


datasketches/src/hll/mod.rs
```diff
@@ -209,14 +209,14 @@ impl Coupon {
...
     /// implementations require a specific value hashing strategy.
     ///
-    /// Hashes `v` using MurmurHash3 128-bit and packs the result into a coupon:
+    /// Hashes `value` using MurmurHash3 128-bit and packs the result into a coupon:
     /// the low 26 bits of the low hash word become the slot index, and the
     /// leading-zero count of the high hash word (capped at 62, then plus one)
     /// becomes the 6-bit register value.
     #[inline(always)]
-    pub fn from_hash<T: Hash>(v: T) -> Self {
+    pub fn from_value<T: Hash>(value: T) -> Self {
```
tisonkun, line 217: > Renames `Coupon::from_hash(value)` to `Coupon::from_value(value)`. The method hashes its generic input internally, so from_hash incorrectly implied that callers supplied an already-computed hash.

cc @notfilippo is this rename reasonable?

ariesdevil (approved):
Semantics are clearer.

ZENOTME (approved):

tisonkun:
Merging ...

@notfilippo your feedback is always welcome since we're still a bit away from a 0.4 release.

datasketches/src/hll/mod.rs
notfilippo: Much clearer. Thanks!

notfilippo (commented):
Makes sense. Thanks!

PR #168, opened 2026-07-28
test: go through crate tests

tisonkun:
See commits for details.


tisonkun:
### Test migration review

I re-reviewed `fa98dce`, `95197af`, and `a93e3c9` against one boundary: tests should protect user-visible behavior or a concrete regression, not private storage layout or control flow.

**Motivation**

- Theta and Tuple use the same generic hash-table machinery, so parallel private suites duplicated coverage and coupled tests to resize and capacity details.
- Malformed-input behavior belongs at the public deserialization boundary, but coverage should remain tied to concrete existing cases rather than expanding into generic prefix enumeration.
- Tuple compaction and generic summary policies are public API capabilities, so their meaningful contracts must remain covered after the internal tests are reduced.

**Result**

- Kept the duplicate private hash-table tests removed. Follow-up `5f27d94` preserves the important selection invariant through the public Theta API: after `trim`, the sketch retains exactly the smallest `k` hashes and sets theta to the next hash. Private table size, probing, resize, and memory-layout checks remain intentionally untested.
- Moved the existing Theta and Tuple malformed-input cases to `serde_tests`, while preserving their concrete truncation scenarios: one missing trailing byte for Theta and a partially truncated final summary for Tuple. Follow-up `4a77bb8` removes the generalized all-prefix helper and six blanket truncation cases that were not traceable to existing regressions; `6f65c94` is therefore superseded by this narrowing.
- Follow-up `863b167` removes the remaining test module from `src/tuple/sketch.rs` and covers compaction through the public integration suite in both exact and estimation modes, including entries, summaries, theta, seed, estimate, and requested ordering. It also restores public generic-policy coverage for a distinct `AddAssign` RHS and for one policy accepting multiple update representations. The narrower max-policy test was replaced rather than duplicated.
- No production behavior was changed.

**Validation**

- `cargo x check`
- `cargo x test` — 475 tests passed, including 125 doctests
- `cargo x lint`

tisonkun:
cc @ariesdevil

tools/download_serde_tests_data.py
tisonkun, line 1: cc @PsiACE I suppose this script can now be a subcommand of `cargo x`.

datasketches/tests/bloom_test/sketch.rs
```diff
@@ -0,0 +1,162 @@
...
+    right.insert("right");
+
+    let left_bits = left.bits_used();
+    let right_bits = right.bits_used();
+
+    let mut intersection = left.clone();
+    intersection.intersect(&right);
+    assert!(intersection.contains(&"shared"));
+    assert!(intersection.bits_used() < left_bits);
+    assert!(intersection.bits_used() <= right_bits);
```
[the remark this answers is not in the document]
tisonkun: Agreed. Since intersection is a bitwise AND, the stable contract is only that the resulting set-bit count cannot exceed either input; strict reduction depended on the incidental bit distribution of these test values. Fixed in `3735ac7` by caching the intersection count and asserting `<=` against both inputs. `cargo x check`, `cargo x test`, and `cargo x lint` all pass.

datasketches/tests/tuple_test/a_not_b.rs
```diff
@@ -15,228 +15,155 @@
...
+        b.update(value, 1u64);
+    }
 
-    let a_not_b = TupleANotB::default();
-    let r = a_not_b
+    let op = TupleANotB::default();
+    let unordered = op.compute(&a, &b, true).unwrap();
+    let ordered = op
         .compute(&a.compact(true), &b.compact(true), true)
         .unwrap();
```
ariesdevil: Seems we also need to test `ordered=false` here.

datasketches/tests/tuple_test/sketch.rs
```diff
@@ -17,284 +17,188 @@
...
-        .sampling_probability(0.5)
-        .build();
-
-    for i in 0..1000 {
-        sketch.update(i, 1u64);
+fn compact_preserves_state_in_exact_and_estimation_modes() {
+    for (lg_k, num_updates, expected_estimation_mode) in [(12, 2_000, false), (5, 5_000, true)] {
+        let mut sketch = default_tuple_sketch_builder().lg_k(lg_k).build();
+        for value in 0..num_updates {
+            sketch.update(value, 1u64);
```
ariesdevil: If we use `1u64`, all the keys are always `1u64`, so we cannot ensure compaction is correct.

```rust
for key in 0..num_updates {
    sketch.update(key, key + 1);
    sketch.update(key, 10);
}
```
I think this pattern will be better.

datasketches/tests/tuple_test/a_not_b.rs
tisonkun: OK. Is this a regression?

I can ask Codex to add it but I'm also verifying if it works unexpectedly.

I glanced every commit but I can miss something concrete.

datasketches/tests/tuple_test/sketch.rs
tisonkun: Great! I get this point.

ariesdevil (approved):
rest LGTM

tisonkun:
@ariesdevil Thanks for your review!

datasketches/tests/tuple_test/sketch.rs
tisonkun: Agreed. Fixed in `e10be7e`: every key now receives `key + 1` and `10`, so summaries vary by key and exercise accumulation. The compact comparison therefore verifies the hash-summary association in both exact and estimation modes.

datasketches/tests/tuple_test/a_not_b.rs
tisonkun: Good catch. The public result-ordering contract had been dropped during consolidation. Restored in `c6579ff`: the test now exercises both `ordered=true` and `ordered=false`, verifies `is_ordered()` follows the request, and confirms theta and retained entries are unchanged.

tisonkun:
@ariesdevil You may review the new commits now :D

ariesdevil (approved):
LGTM again:)

ZENOTME (approved):
Thanks @tisonkun !

PR #169, opened 2026-07-28
test: move test data preparation into xtask

PsiACE:
# Which issue does this PR close?

N/A.

# Rationale for this change

Preparing cross-language serialization test data currently depends on a standalone Python script. Moving this workflow into xtask gives contributors and CI a single project-native command.

# What changes are included in this PR?

- Add `cargo x test-data` to download the pinned DataSketches TCK archive and synchronize the C++ and Java snapshots.
- Update CI, contributor documentation, and missing-data guidance to use the new command.
- Remove `tools/download_serde_tests_data.py`.
- Preserve stale snapshot cleanup and atomic file replacement.

Validated with `cargo x check`, `cargo x test-data`, and `cargo x test`, including all 52 serialization compatibility tests.

# Are there any user-facing changes?

Yes. Contributors should now run `cargo x test-data` instead of the Python script. There are no changes to the public library API.

# AI Usage Statement

This PR was developed with assistance from OpenAI Codex (GPT-5) for implementation and review.


tisonkun (commented):
Generally OK.

But the command name should not be `test-data` at least not alone.

Why `test-data` means download serialization test snapshots from another repo?

PsiACE:
we can just rename this command to `download-test-data` or `prepare-test-data`

tisonkun (commented):
Yeah. I think we can call it `prepare-testdata` and have an argument like the original script (`--java`, `--cpp`, `--all`).

We may have `--go` to fetch Go snapshots and `--rust` to build Rust snapshots with a generator binary (this is why we call it "preapre" rather than "download").

tisonkun (commented):
LGTM. I'll pull this patch and merge it later today.

tisonkun (approved):
Thanks!

PR #172, opened 2026-07-28
test: add cross-language regression coverage

tisonkun:
## Summary

- exercise Theta A-not-B when the larger B input contributes the lower theta, for both ordered and unordered paths
- retain an explicit two-sketch HLL commutativity check with exact estimate equality, and add the reference merge-order regression across all six permutations
- replay the original Java and C++ CPC fixture inputs after deserialization to verify mutable state across sparse, hybrid, pinned, and sliding flavors

## Motivation

The Java and C++ implementations protect these observable behaviors with focused regression coverage. Rust already produces the expected results, but its existing tests either used a less discriminating input shape or stopped at serialization round-trip checks. This PR strengthens those public boundaries without copying language-specific memory, allocator, or internal-state tests.

Each test concern is isolated in its own commit.

## Validation

- cargo x check
- cargo x test
- cargo x lint

### DIFF
--- a/datasketches/tests/serde_tests/cpc.rs
+++ b/datasketches/tests/serde_tests/cpc.rs
@@ -16,7 +16,7 @@
 // under the License.
 
 use std::fs;
-use std::path::PathBuf;
+use std::path::Path;
 
 use datasketches::cpc::CpcSketch;
 use googletest::assert_that;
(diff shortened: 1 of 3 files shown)


datasketches/tests/theta_test/a_not_b.rs
tisonkun, line 1: This is reviewed by me.

datasketches/tests/serde_tests/cpc.rs
tisonkun, line 1: This is reviewed by me. But perhaps we can tidy the code a bit.

tisonkun: Fixed.

PsiACE (approved):
totally lgtm

Issue #137, opened 2026-06-09
Tracking issue for reporting sketches estimated size


tisonkun:
I think the `estimated_` contract is fine. I think in a follow-up we could also introduce tracking of correct size for sketches that can report correct sizing information.

_Originally posted by @notfilippo in https://github.com/apache/datasketches-rust/pull/136#pullrequestreview-4449930212_

Renkai:
I it a good first issue? I will try work on this.

tisonkun:
@Renkai Your patch has been merged.

May you check if all sketches have an `estimated_size` method now and thus we may close this issue as resolved.

Renkai:
Hi, @tisonkun 
For basic sketches, I think they have all included. But there are set operations, which behave similar to sketches in some way, I'm not sure if they shall include this method. If needed, there is a draft PR: https://github.com/apache/datasketches-rust/pull/177/changes

tisonkun:
Closed as existing sketches are all covered now.

New sketches can decide whether or not to provide a estimated size method when introduced.

PR #174, opened 2026-08-01
feat: expose estimated_size for countmin and frequencies sketches

Renkai:
This is a follow-up of #136, addressing part of #137.

Add `estimated_size()` to the remaining sketches that don't report their memory footprint yet:

- `CountMinSketch`: `size_of::<Self>()` plus the capacity of the `counts` and `hash_seeds` backing vectors
- `FrequentItemsSketch`: `size_of::<Self>()` plus the heap allocations of the internal `ReversePurgeItemHashMap` (keys/values/states arrays)

The implementation follows the pattern established in #136: heap usage is measured by allocation capacity, and nested storage types expose their own `estimated_size()` helper.

Tests: added one case per sketch (`countmin_test`, `frequencies_test`). Verified with `cargo x check`, `cargo x test`, and `cargo x lint`.


tisonkun (approved):
Thanks for your contribution!

PR #177, opened 2026-08-02
feat: expose estimated_size for stateful set operations

Renkai:
Follow-up of #174, addressing part of #137.

While all sketch types expose `estimated_size()` now, the stateful set-operation operators — which hold real heap state while accumulating updates — don't report their footprint yet. This patch adds `estimated_size()` to:

- `HllUnion` — delegates to its internal `HllSketch` gadget
- `CpcUnion` — covers both internal states (`Accumulator` sketch and `BitMatrix`)
- `ThetaUnion` / `ThetaIntersection` — via a heap-only helper on `UnionState` / `IntersectionState`
- `TupleUnion` / `TupleIntersection` — same shared helpers

The a-not-b operators are stateless (a one-shot computation over two inputs), so they are intentionally exempt.

Implementation notes:

- `UnionState::estimated_size()` / `IntersectionState::estimated_size()` report heap allocations only (delegating to `SketchHashTable::estimated_size()`), so the public operators can compose `size_of::<Self>() + state.estimated_size()` without double counting.
- For `HllUnion` and `CpcUnion`, whose state embeds a full sketch inline, the inner sketch's inline size is subtracted for the same reason.

Note that an operator's footprint cannot be inferred from its result sketch: the internal table keeps capacity headroom per its load factor and resize history (and `CpcUnion`'s bit matrix is fixed at `k * 8` bytes), whereas `to_sketch()` produces a tightly packed copy.

Tests: one case per operator in the existing integration targets. Verified with `cargo x check`, `cargo x test` (incl. Go snapshots), and `cargo x lint`.


datasketches/tests/cpc_test/union.rs
```diff
@@ -179,3 +179,17 @@ fn test_lg_k_too_small() {
...
+    let empty_size = union.estimated_size();
+    assert!(empty_size > 0);
+
+    let mut sketch = CpcSketch::new(11);
+    for i in 0..1000 {
+        sketch.update(i);
+    }
+    union.update(&sketch);
+    assert!(union.estimated_size() > empty_size);
+}
```
tisonkun, line 194: Please try to follow https://github.com/apache/datasketches-rust/pull/174/changes/05b5ad519327ede5088cacab12b9382d9553f62f to assert exact values of estimated_size.

tisonkun: Ditto other similar tests.

tisonkun (commented):
Thanks for your contribution @Renkai!

Looks like a good suppliment.

cc @notfilippo @ariesdevil FYI you may review how HllSktech and ThetaSketch/TupleSketch provides their estimated size.

Comments inline above.

Renkai:
updated

tisonkun (approved):
Thanks!

PR #178, opened 2026-08-02
refactor: reorganize theta family mods and add key-only sketch views

ariesdevil:
## Summary

- Group the shared Theta, Theta, and Tuple implementations under `theta_family`.
  - Preserve the existing public module paths through re-exports.
- Introduce key-only sketch views for operations that only consume retained hashes.
  - Update Jaccard similarity and the B side of A-not-B to avoid cloning Tuple summaries.
  - Keep full retained-entry views for operations such as union and intersection.

The theta family becomes:
```
datasketches/src/theta_family/
├── common/
├── theta/
├── tuple/
└── mod.rs
```


tisonkun (approved):
Generally LGTM.

But I'd still prefer:

```
theta
thetacommon
tuple
```

.. structure.

But I can submit a follow-up PR to demonstrate that structure tomorrow.

tisonkun:
Now I know you use a `theta_family` mod to keep several methods `pub(super)`.

tisonkun:
I pushed https://github.com/apache/datasketches-rust/pull/179 to rename `theta_family` to `thetafamily`. It's not exported so it makes no difference from user interfaces, but keep our top-level modules no separator which provides a stable style (my bias, yes).

ariesdevil:
> But I'd still prefer:
> 
> ```
> theta
> thetacommon
> tuple
> ```
> 
> .. structure.

Actually, I like the original structure too, but I don't want to put `thetacommon` at the top level, so I came up with a compromise.

PR #179, opened 2026-08-02
refactor: simplify hash and theta module layout

tisonkun:
## Summary

- rename the internal theta_family module to thetafamily
- move hash_value wrappers under hash::value while preserving their value-specific documentation
- consolidate binomial proportion bounds into binomial_bounds

## Testing

- cargo x check
- cargo x test
- cargo x lint


tisonkun:
cc @ariesdevil

PsiACE (approved):

PR #181, opened 2026-08-03
feat: add exact equality for Jaccard sketches

hawkingrei:
## What changed

- Add `exactly_equal` to `ThetaJaccardSimilarity` and `TupleJaccardSimilarity`.
- Reuse the key-only union path to compare retained hashes and theta.
- Keep Tuple summary values and types outside the equality comparison.
- Cover empty, mutable/compact, seed mismatch, summary-independent, and zero-retained cases.

## Why

This is a focused follow-up to #142 and its final review feedback. It adds only exact sketch-state equality; `similarity_test` and `dissimilarity_test` remain separate future work.

The zero-retained case intentionally distinguishes the two APIs: `compute` preserves an uncertain Jaccard interval, while `exactly_equal` compares the observable sketch state (retained hashes and theta), matching the other DataSketches implementations.

## Related

- Follow-up to #142

## Testing

- `cargo x prepare-testdata`
- `cargo x check`
- `cargo x test`
- `cargo x lint`

### DIFF
--- a/datasketches/src/thetafamily/common/jaccard_similarity.rs
+++ b/datasketches/src/thetafamily/common/jaccard_similarity.rs
@@ -25,6 +25,7 @@ use crate::thetacommon::binomial_bounds;
 use crate::thetacommon::constants::MAX_LG_K;
 use crate::thetacommon::constants::MAX_THETA;
 use crate::thetacommon::constants::MIN_LG_K;
+use crate::thetacommon::hash_table::CompactSketchParts;
 use crate::thetacommon::intersection::IntersectionMergePolicy;
 use crate::thetacommon::intersection::IntersectionState;
 use crate::thetacommon::union::UnionMergePolicy;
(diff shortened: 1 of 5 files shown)


datasketches/src/thetafamily/common/jaccard_similarity.rs
```diff
@@ -299,6 +269,71 @@ impl JaccardSimilarityOperator {
...
+
+    pub(crate) fn exactly_equal<A, B>(&self, sketch_a: &A, sketch_b: &B) -> Result<bool, Error>
+    where
+        A: ThetaKeySketchView,
+        B: ThetaKeySketchView,
+    {
+        if sketch_a.is_empty() && sketch_b.is_empty() {
+            return Ok(true);
+        }
+        if sketch_a.is_empty() || sketch_b.is_empty() {
```
ariesdevil, line 281: `exactly_equal` returns `Ok(false)` as soon as exactly one sketch is empty, before calling `compute_union`, where seed validation happens.
But the docs in `datasketches/src/thetafamily/theta/jaccard_similarity.rs:84` and `datasketches/src/thetafamily/tuple/jaccard_similarity.rs:90` say this will return an error.

ariesdevil:
rest LGTM

datasketches/src/thetafamily/common/jaccard_similarity.rs
hawkingrei: Thanks for catching this. The empty-input short-circuit is intentional, so I updated both public docs to state that seed mismatches are errors only when both sketches are non-empty. I also added Theta and Tuple regression coverage for an empty sketch paired with a non-empty sketch built with a different seed.

datasketches/src/thetafamily/common/jaccard_similarity.rs
```diff
@@ -299,6 +269,71 @@ impl JaccardSimilarityOperator {
...
+                "incompatible seed hash: expected {}, got {}",
+                expected,
+                sketch.seed_hash(),
+            )));
+        }
+        Ok(())
+    }
+}
+
+fn identical_sets<A, B>(sketch_a: &A, sketch_b: &B, union: &CompactSketchParts<KeyEntry>) -> bool
```
ariesdevil, line 332: This function name is a little bit confuse, add some docs or renaming to `same_retained_keys_and_theta` seems better.

ariesdevil (approved):
LGTM

datasketches/src/thetafamily/common/jaccard_similarity.rs
tisonkun: @hawkingrei @ariesdevil Is the real manner from code what other DataSketches impls do?

ariesdevil: yup, https://github.com/apache/datasketches-cpp/blob/c22888581964fc490feee835766cc8c3adb722e0/theta/include/theta_jaccard_similarity_base.hpp#L85

datasketches/src/thetafamily/common/jaccard_similarity.rs
ariesdevil: Oh, I see the C++ impl uses the same name, so just adding some docs is fine for me.

tisonkun (approved):

datasketches/src/thetafamily/common/jaccard_similarity.rs
tisonkun: @ariesdevil You can use the "suggestion" feature to add a suggested comment here - https://docs.github.com/en/pull-requests/how-tos/review-pull-requests/incorporating-feedback-in-your-pull-request

hawkingrei: Thanks. I kept the C++-aligned name and added documentation in f67e857 explaining why no additional retained union keys plus matching theta values proves that both sketches have the same retained key set and theta.

PR #183, opened 2026-08-05
refactor: centralize seed hash validation

### LINKED ISSUE #182: Check seed_hash can be a helper function
We check seed_hash's equality in several places, and it would be better to have a helper for the repeated logic.

A minor benefit is that it can reduce duplicate seed_hash calculations when it is. But that means an error code path, so it doesn't matter that much.

jaideeppyne:
Closes #182.

## What changed

- add one internal `check_seed_hash` helper that computes the expected hash once and preserves the existing invalid-data error
- use it in the Count-Min, CPC, compact Theta (v1/v2/v3), and compact Tuple deserialization paths
- remove 41 lines of repeated validation while keeping the helper crate-private

## Validation

- `cargo x prepare-testdata` — passed (645 cross-language snapshots prepared)
- `cargo x test` — passed
- nightly Clippy with `-D warnings` — passed
- nightly rustfmt check — passed
- rustdoc with warnings denied — passed
- Taplo format check and typos — passed

`cargo x lint` reached and passed all repository/code checks above, but its final local-tool bootstrap could not install `hawkeye 6.5.1` with the repository's Rust 1.86 toolchain because current transitive crates require Rust 1.88+; no source/license finding was reported before that environment-only bootstrap failure.

## Risk and rollback

This is an internal refactor with no public API or serialized-format change. The main risk is an accidentally changed error kind/message; the helper retains `ErrorKind::InvalidData` through `Error::deserial` and the prior message text. Reverting commit `0d6570b` fully rolls it back.

## AI assistance

OpenAI Codex assisted with implementation and validation. I reviewed the diff and test results.

### DIFF
--- a/datasketches/src/countmin/mod.rs
+++ b/datasketches/src/countmin/mod.rs
@@ -15,9 +15,9 @@
 // specific language governing permissions and limitations
 // under the License.
 
-//! Count-Min sketch implementation for frequency estimation.
+//! CountMin sketch implementation for frequency estimation.
 //!
-//! The Count-Min sketch provides approximate frequency counts for streaming data
+//! The CountMin sketch provides approximate frequency counts for streaming data
 //! with configurable relative error and confidence bounds.
 //!
 //! # Usage
--- a/datasketches/src/countmin/sketch.rs
+++ b/datasketches/src/countmin/sketch.rs
@@ -377,7 +377,11 @@ impl<T: CountMinValue> CountMinSketch<T> {
             .map_err(insufficient_data("seed_hash"))?;
         cursor.read_u8().map_err(insufficient_data("unused8"))?;
 
-        check_seed_hash(seed_hash, seed)?;
+        check_seed_hash(compute_seed_hash(seed), seed_hash, |expected, actual| {
+            Error::deserial(format!(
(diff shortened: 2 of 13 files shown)


tisonkun (changes requested):
Different checks may return different error.

Perhaps a check_seed_hash(expect, actual, |expect, actual|...) help. I don't know.

jaideeppyne:
Thanks — updated in 6b193ad. `check_seed_hash` now accepts an error-producing closure over `(expected, actual)`, so each caller retains control of its error contract. In particular, the CPC path still uses its prior explicit `Error::new(ErrorKind::InvalidData, ...)` form, while the deserialization paths retain `Error::deserial(...)`; the helper only owns the equality check. This also keeps the mismatch path to one seed-hash computation.

Validation after the change:
- `cargo x test` — passed
- `cargo +nightly clippy --tests --all-features --all-targets --workspace -- -D warnings` — passed
- `cargo +nightly fmt --all --check` — passed
- `RUSTDOCFLAGS="--cfg docsrs -D warnings" cargo +nightly doc --package datasketches --all-features --no-deps` — passed

tisonkun (approved):
Fixed up. LGTM.

PR #185, opened 2026-08-09
refactor: replace theta family sketch traits with concrete views

tisonkun:
## Summary

- replace the publicly implementable Theta-family sketch traits with concrete, cheap-copy `ThetaSketchView<'a>` and `TupleSketchView<'a, S>` values backed by private mutable/compact state
- preserve caller ergonomics through `From<&...>` / `Into<...>` conversions and `as_view()`, while returning `impl Iterator` from view iteration so the iterator state remains private
- pass complete sketch views into shared union, intersection, A-not-B, and Jaccard implementations instead of splitting one sketch into scalar values and iterators at each call site
- keep code reuse behind Theta-family-only capabilities (`SketchEntry`, `KeySketch`, `EntrySketch`, and `SketchScalars`) rather than exposing an extension point to downstream crates
- simplify A-not-B and Jaccard by caching seed state in their public wrappers and delegating to internal free functions, removing cache-only common operator types
- define internal visibility at module boundaries and remove `TestEntry` / `TestSketch` white-box adapters in favor of the Theta and Tuple behavior suites

## Rationale

Theta-family set operations accept a closed set of representations: mutable and compact Theta or Tuple sketches. The previous public traits exposed seed, theta, ordering, retained-entry, and iterator invariants as a downstream implementation contract even though arbitrary external sketch implementations were not a supported extension point.

The concrete views make the supported representations explicit without changing the common call pattern: references to mutable and compact sketches can still be passed directly, and callers can request an explicit view with `as_view()`. Tuple key-only operations can inspect hashes without requiring summaries to implement `Clone`; operations that retain summaries express the `Clone` requirement only where entries must be materialized.

Shared algorithms still use static dispatch, but their capability traits and scalar bundle are confined to `thetafamily`. Passing a complete view keeps each sketch state together and lets the shared implementation select the scalar values, hashes, or entries it needs without leaking those internal abstractions into the public API.

This revisits the abstraction discussed in #151. This closes #151.

## Testing

- `cargo x check`
- `cargo x test`
- `cargo x lint`

### DIFF
--- a/datasketches/src/thetafamily/common/a_not_b.rs
+++ b/datasketches/src/thetafamily/common/a_not_b.rs
@@ -33,7 +33,7 @@ use crate::thetacommon::hash_table::CompactSketchParts;
 /// Surviving entries are moved from `A` unchanged, and `B` contributes only hashes, so unlike
 /// the union and intersection this operation needs neither matching entry types nor an
 /// entry-merge policy.
-pub(in crate::thetafamily) fn compute<A, B>(
+pub fn compute<A, B>(
     seed_hash: u16,
     a: A,
     b: B,
--- a/datasketches/src/thetafamily/common/a_not_b.rs
+++ b/datasketches/src/thetafamily/common/a_not_b.rs
@@ -21,8 +21,10 @@ use crate::error::Error;
 use crate::error::ErrorKind;
 use crate::hash::check_seed_hash;
 use crate::hash::compute_seed_hash;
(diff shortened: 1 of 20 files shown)


tisonkun:
cc @ariesdevil

ZENOTME (approved):
Nice work

ariesdevil (commented):
Nice work

Issue #176, opened 2026-08-01
Rust Snapshot Generator


tisonkun:
This issue supersedes https://github.com/apache/datasketches-rust/issues/10 to squash the historical discussion.

Currently, the C++/Go/Java implementations of DataSketches provide snapshot generators for sketches' serialization data.

You can check the following scripts for how their snapshots get generated:

* https://github.com/apache/datasketches-tck/blob/d363b12d/serialization/cpp/generator.py
* https://github.com/apache/datasketches-tck/blob/d363b12d/serialization/go/generator.py
* https://github.com/apache/datasketches-tck/blob/d363b12d/serialization/java/generator.py

We may provide a binary in datasketches-rust as well to generate snapshots for supported sketches' serialization data.

I can imagine it as a local workspace member depending on `datasketches` and we will have a `cargo x` task to invoke the generator for automation. Or else, being part of the `serde_tests` but write the Rust sketch's serialization data to a specific location.

tomatotomata:
I was thinking of keeping this as a small workspace member under `xtask` that depends on `datasketches`, writes deterministic per-sketch snapshots into a dedicated output directory, and exposes one `cargo x` entry point for the existing serialization workflow. I would mirror the TCK generators only for the sketches currently covered by `serde_tests`, add a focused CLI test, and keep generated files out of Git as the contributor guide describes. Does that scope fit the intended first step for #176?

tisonkun:
@tomatotomata Please go ahead. While I don't think it would be a "small workspace member under xtask", but a workspace member at the same level to `datasketches`, depending on `datasketches` and being called by `xtask` via cargo comands, you can submit a PR so that our discussion would be more concrete.

PR #189, opened 2026-08-09
fix: preserve non-empty theta v2 exact sketches

### LINKED ISSUE #186: Theta v2 exact sketches deserialize as logically empty
## Problem

`CompactThetaSketch::deserialize_v2` reads the retained entries from a version 2 exact compact image (`preamble_longs == 2`) but always constructs the sketch with `empty: true`.

A valid v2 exact image with one retained hash therefore deserializes with all of the following contradictory observations:

- `num_retained() == 1`
- `is_empty() == true`
- `estimate() == 0.0`

Besides returning the wrong estimate, the false logical-empty state can cause set operations to skip a real input.

Reviewed revision: `4266ad9cad21e0660c2ae9506174e01665bd100d`.

## Expected behavior

For a v2 exact image (`preamble_longs == 2`), logical emptiness is determined by `num_entries == 0`. An image with one or more retained entries must be non-empty and estimate its exact retained count.

## Relationship to other implementations

The current C++ v2 reader implements this rule by returning empty only when `num_entries == 0`, and otherwise constructs a non-empty compact sketch:

https://github.com/apache/datasketches-cpp/blob/c22888581964fc490feee835766cc8c3adb722e0/theta/include/theta_sketch_impl.hpp#L605-L631

The current Go v2 decoder uses the same rule:

https://github.com/apache/datasketches-go/blob/c558cc2d64f9a307c196a7adb0067eadd1b776f7/theta/decoder.go#L295-L343

Go also retains compatibility fixtures explicitly identified as Java-generated v2 compact images. Current Java only heapifies compact serialization versions 3 and 4, so C++ and Go provide the active reference behavior for reading legacy Java v2 images:

https://github.com/apache/datasketches-go/blob/c558cc2d64f9a307c196a7adb0067eadd1b776f7/theta/sketch_serialization_test.go#L210-L228

https://github.com/apache/datasketches-java/blob/4067ffefabbcd03944dc3617ff2948ab760f3b75/src/main/java/org/apache/datasketches/theta/CompactThetaSketch.java#L97-L115

## Proposed fix

Set `empty` from `num_entries == 0` in the v2 exact branch, matching the current C++ and Go legacy readers.

## Acceptance criteria

- Add a regression test for a v2 exact image with retained entries and verify retained count, logical emptiness, estimate, and bounds.
- Add or retain coverage showing a zero-entry v2 exact image remains empty.
- Verify the decoded sketch survives a current-format round trip without losing its entries or estimate.
- Run the repository check, test, and lint workflows.

tisonkun:
## Summary

Fix legacy Theta serialization version 2 exact images being deserialized as logically empty when they contain retained entries.

The v2 exact branch now derives logical emptiness from `num_entries == 0` instead of setting it unconditionally.

Closes #186.

## Regression coverage

The new serialization compatibility tests construct valid v2 exact images using the current Theta family ID and default seed hash, then verify:

- a non-empty image retains all entries,
- `is_empty()` is false,
- exact estimate and bounds equal the retained count,
- entry iteration is unchanged,
- a current-format round trip preserves the state, and
- a zero-entry v2 exact image remains empty.

## Relationship to other implementations

This restores agreement with the current C++ and Go legacy readers. Both treat a v2 exact image (`preamble_longs == 2`) as empty only when `num_entries == 0`:

- C++: https://github.com/apache/datasketches-cpp/blob/c22888581964fc490feee835766cc8c3adb722e0/theta/include/theta_sketch_impl.hpp#L605-L631
- Go: https://github.com/apache/datasketches-go/blob/c558cc2d64f9a307c196a7adb0067eadd1b776f7/theta/decoder.go#L295-L343

Go retains compatibility fixtures identified as Java-generated v2 images. Current Java reads compact versions 3 and 4 rather than v2 directly, so the active legacy-reader behavior is represented by C++ and Go:

- Go Java-v2 fixture test: https://github.com/apache/datasketches-go/blob/c558cc2d64f9a307c196a7adb0067eadd1b776f7/theta/sketch_serialization_test.go#L210-L228
- Current Java version dispatch: https://github.com/apache/datasketches-java/blob/4067ffefabbcd03944dc3617ff2948ab760f3b75/src/main/java/org/apache/datasketches/theta/CompactThetaSketch.java#L97-L115

The fix changes no serialization bytes emitted by Rust; it only corrects the state reconstructed from legacy input.

## Validation

- `cargo x prepare-testdata`
- `cargo x check`
- `cargo x test`
- `cargo x lint`

### DIFF
--- a/datasketches/tests/serde_tests/theta.rs
+++ b/datasketches/tests/serde_tests/theta.rs
@@ -18,6 +18,8 @@
 use std::fs;
 use std::path::PathBuf;
 
+use datasketches::codec::SketchBytes;
+use datasketches::common::NumStdDev;
 use datasketches::error::ErrorKind;
 use datasketches::theta::CompactThetaSketch;
 use datasketches::theta::ThetaSketchBuilder;
@@ -25,6 +27,24 @@ use googletest::assert_that;
 use googletest::prelude::near;
 
 use crate::serialization_test_data;
+
+fn serialize_v2_exact(entries: &[u64]) -> Vec<u8> {
+    let current = ThetaSketchBuilder::default().build().compact(true);
+    let current_bytes = current.serialize();
+    let mut bytes = SketchBytes::with_capacity((2 + entries.len()) * size_of::<u64>());
+    bytes.write_u8(2); // preamble longs
+    bytes.write_u8(2); // serialization version
+    bytes.write_u8(current_bytes[2]); // theta family ID
+    bytes.write_u8(0); // unused
+    bytes.write_u16_le(0); // unused
+    bytes.write_u16_le(current.seed_hash());
+    bytes.write_u32_le(entries.len() as u32);
+    bytes.write_u32_le(0); // unused
+    for &entry in entries {
+        bytes.write_u64_le(entry);
+    }
+    bytes.into_bytes()
+}
 
 fn test_sketch_file(path: PathBuf, expected_cardinality: usize, use_compressed_round_trip: bool) {
     let expected = expected_cardinality as f64;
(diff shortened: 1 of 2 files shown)


datasketches/tests/serde_tests/theta.rs
```diff
@@ -26,6 +28,24 @@ use googletest::prelude::near;
 
 use crate::serialization_test_data;
 
+fn serialize_v2_exact(entries: &[u64]) -> Vec<u8> {
```
ZENOTME, line 31: I'm not sure there are better way to provide a v2 serialize snapshot instead of construct by hand here. 🤔

ZENOTME (approved):
LGTM. Maybe we refine test later.

datasketches/tests/serde_tests/theta.rs
tisonkun: > v2 serialize snapshot

Maybe. But existing C++/Java/Go impls do not provide one. We may provide one when implementing #176 , and if you'd like to, you can go to other impls to ask them to provide one and update https://github.com/apache/datasketches-tck.

Issue #194, opened 2026-08-11
Clarify BloomFilter semantics after invert and correct builder documentation


tisonkun:
Three related BloomFilter contract/documentation issues should be resolved together:

1. invert flips the backing bit array and therefore the resulting object no longer satisfies the normal Bloom-filter no-false-negative guarantee advertised at the type and module level.
2. is_empty is documented in terms of inserted items, but after invert it only reports whether the backing bit array has zero set bits.
3. BloomFilterBuilder::build documents a panic when neither accuracy nor size was configured, but that state is not constructible through the public API: the builder has no public default constructor and both public entry points initialize a valid configuration.

Design questions:

- Should invert remain on the normal BloomFilter API, move to a raw/diagnostic API, or make the altered guarantee observable in the type/state?
- Should is_empty retain a logical-stream meaning, be renamed to describe bit state, or document mode-dependent semantics?
- Which module/type/method guarantees need a post-invert qualification?
- Remove the unreachable build panic documentation and, if possible, encode the configured builder state more directly.

This issue intentionally excludes adding try_union/try_intersect; that is a separate fallibility/API-consistency decision.

PR #196, opened 2026-08-11
docs: correct public sketch semantics

tisonkun:
## Summary

- correct Bloom membership examples: positive results mean possibly present, while negative results mean definitely absent in normal operation
- document Hll4, Hll6, and Hll8 as isomorphic register representations with identical accuracy for the same lg_k
- explain how HLL union coupon-mode replay and array-mode downsampling determine the result's effective lg_k and error distribution
- correct the documented widths of the four 64-bit SketchSlice readers
- clarify that core Frequencies operations require Eq + Hash, while FrequentItemValue is required only for built-in serialization
- align Theta and Tuple Jaccard seed-error documentation with the empty-input fast paths

The broader post-invert Bloom contract remains tracked in #194.

## Validation

- cargo x check
- cargo x test
- cargo x lint

### DIFF
--- a/datasketches/src/frequencies/mod.rs
+++ b/datasketches/src/frequencies/mod.rs
@@ -25,9 +25,7 @@
 //!
 //! This sketch tracks approximate frequencies of items of type `T` that implement [`Eq`] and
 //! [`Hash`](std::hash::Hash), with optional associated counts (`T` item, `u64` count) that are
-//! members of a multiset. The true frequency of an item is the sum of its associated counts. Core
-//! updates and queries do not require [`FrequentItemValue`]; that trait is required only by the
-//! built-in serialization and deserialization methods.
+//! members of a multiset. The true frequency of an item is the sum of its associated counts.
 //!
 //! This implementation provides the following capabilities:
 //! * Estimate the frequency of an item.
@@ -91,8 +89,7 @@
 //! # Serialization
 //!
 //! The built-in serialization methods are available when the item type implements
-//! [`FrequentItemValue`]. Sketches whose items implement only [`Eq`] and [`Hash`](std::hash::Hash)
-//! can still use all core update and query operations.
+//! [`FrequentItemValue`].
 //!
 //! ```
 //! use datasketches::frequencies::FrequentItemsSketch;
--- a/datasketches/src/frequencies/mod.rs
+++ b/datasketches/src/frequencies/mod.rs
@@ -23,10 +23,11 @@
 //! in Data Streams"](https://arxiv.org/abs/1705.07001) by Daniel Anderson, Pryce Bevan, Kevin Lang,
 //! Edo Liberty, Lee Rhodes, and Justin Thaler.
 //!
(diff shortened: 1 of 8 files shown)


datasketches/src/hll/mod.rs
```diff
@@ -42,24 +42,29 @@
...
+//! The three target HLL types are isomorphic representations of the same registers. Given the
+//! same `lg_k` and input, they produce identical estimates and error distributions; they trade
+//! memory layout and update performance, not statistical precision:
 //!
-//! * [`HllType::Hll4`]: 4 bits per bucket (most compact)
-//! * [`HllType::Hll6`]: 6 bits per bucket (balanced)
-//! * [`HllType::Hll8`]: 8 bits per bucket (highest precision)
+//! * [`HllType::Hll4`]: 4 bits per bucket plus an auxiliary table for rare exceptions (most
+//!   compact)
+//! * [`HllType::Hll6`]: 6 bits per bucket (fixed-size middle ground)
```
tisonkun, line 50:
```suggestion
//! * [`HllType::Hll4`]: 4 bits per bucket (most compact)
```

Need not expose internal concepts.

datasketches/src/frequencies/mod.rs
tisonkun, line 1: Maybe keep each section focused. Need not mention the other part.

datasketches/src/hll/union.rs
```diff
@@ -45,6 +45,12 @@ use crate::hll::mode::Mode;
 /// the union of all input sketches. It automatically handles sketches with
 /// different configurations and modes.
 ///
+/// Coupon-mode inputs are replayed into the current gadget without reducing its `lg_k`. An
+/// array-mode input with a smaller `lg_k` reduces the gadget to that value; larger array inputs are
+/// downsampled to the gadget's current configuration. Once reduced, the effective `lg_k` remains
+/// lower until [`reset`](Self::reset), so estimates and bounds reflect the reduced register count.
+/// The requested [`HllType`] changes only the result representation, not its statistical accuracy.
+///
```
tisonkun, line 51: Is `Coupon-mode` and `array-mode` user facing concept?

datasketches/src/hll/mod.rs
tisonkun: Applied. The Hll4 bullet now stays at the user-visible 4-bit/compact trade-off and no longer mentions the auxiliary table.

datasketches/src/frequencies/mod.rs
tisonkun: Applied. The overview now states only the core Eq + Hash requirement, while the Serialization section states only the FrequentItemValue requirement.

datasketches/src/hll/union.rs
tisonkun: They are internal representation concepts. I removed those terms from both the HllUnion type documentation and the module overview, and kept the contract in terms of the observable effective lg_k and its effect on accuracy.

PR #197, opened 2026-08-11
docs: standardize public API documentation style

tisonkun:
## Summary

- Standardize publicly reachable rustdoc summaries on third-person present tense, using forms such as `Creates`, `Updates`, and `Returns`.
- Normalize sentence punctuation, inline-code formatting for identifiers and ranges, and `lg_k` terminology while retaining exact Rust parameter names where needed.
- Reorder existing sections as Examples, Errors, and Panics, followed by compatibility notes; no failure-semantics sections were added.
- Document these conventions in `CONTRIBUTING.md`.

## Scope

This is a documentation-only refactor. It does not change code, public APIs, or runtime behavior.

## Validation

- `cargo x check`
- `cargo x test`
- `cargo x lint`
- Rustdoc JSON audit across all features for summary voice, punctuation, terminology, and section order

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -60,6 +60,13 @@ Lint:
 cargo x lint
 ```
 
+## Public API documentation
+
+- Describe types with noun phrases and API behavior with third-person present-tense verbs such as `Creates`, `Updates`, and `Returns`.
+- End summary sentences with punctuation, and format Rust identifiers, literals, and numeric ranges as inline code.
+- Use `lg_k` for the algorithm parameter in prose. Use a different name only when referring to an exact Rust identifier or an external serialization format.
+- When applicable, order sections as `# Examples`, `# Errors`, and `# Panics`, followed by compatibility notes. Include only sections that describe an actual contract.
+
 ## Integration test layout
 
 Integration tests for the `datasketches` crate live under `datasketches/tests` and use two entry-point patterns.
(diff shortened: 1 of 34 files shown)


CONTRIBUTING.md
```diff
@@ -60,6 +60,13 @@ Lint:
 cargo x lint
 ```
 
+## Public API documentation
+
+- Describe types with noun phrases and API behavior with third-person present-tense verbs such as `Creates`, `Updates`, and `Returns`.
+- End summary sentences with punctuation, and format Rust identifiers, literals, and numeric ranges as inline code.
+- Use `lg_k` for the algorithm parameter in prose. Use a different name only when referring to an exact Rust identifier or an external serialization format.
```
tisonkun: Why is `lg_k` so special to be mentioned here?

CONTRIBUTING.md
```diff
@@ -60,6 +60,13 @@ Lint:
...
+- When applicable, order sections as `# Examples`, `# Errors`, and `# Panics`, followed by compatibility notes. Include only sections that describe an actual contract.
```
tisonkun: I think `Examples` should go in the last position; that is, the contract first, then examles.

CONTRIBUTING.md
tisonkun: Agreed. lg_k is a one-off terminology cleanup rather than a general documentation rule, so I removed this bullet from CONTRIBUTING.

CONTRIBUTING.md
tisonkun: Agreed. I changed the convention to put contract sections and compatibility notes first, with Examples last, and applied that ordering throughout this PR.

tisonkun:
All reviewed. The documents of `HllType` can be reverted to contain more details. But the current version focuses more on user-facing information and I think it's fair enough to do the change now until someone has a different opinion.

Merging ...

Issue #90, opened 2026-02-13
REQ sketch


pmcgleenon:
Hello 

thanks for creating this project!

There is a rust implementation of Relative Error Quantiles sketch derived from the C++ apache datasketches reference here https://github.com/pmcgleenon/reqsketch-rs

Maybe it could be used as a starting point for a REQ implementation in this project, it would need changes for the serialization and maybe other areas as well as a careful review

What do you think about adapting this existing rust implementation, or do you think a fresh start would be better?

Xuanwo:
Let me take a look.

pmcgleenon:
@Xuanwo  interested in your thoughts. I prototyped adding apache-datasketches compatible serialization to https://github.com/pmcgleenon/reqsketch-rs and it seems straightforward to add this

Do you have any opinions or guidance on the current reqsketch-rs API?    If you want I can create a PR for review

By the way there is a benchmark that compares different types of sketches https://github.com/drmingdrmer/rust-histogram-benchmark, it's using version `0.0.1` of reqsketch-rs.    There have been some improvements since `0.0.1` to align closer with the C++ apache datasketches implementation

tisonkun:
@pmcgleenon Thanks for reaching out!

It seems [reqsketch-rs](https://github.com/pmcgleenon/reqsketch-rs) is inspired by datasketches' impl. Is its interface similar to datasketches'? If so, would you like to upstream your reqsketch to this lib? I'm glad to review any pull request in this direction.

tisonkun:
For example, @andylokandy and I previously made [tdigests](https://github.com/andylokandy/tdigests) but now we prefer this lib's tdigest impl. Similar to @mrcroxx's [cmsketch](https://docs.rs/cmsketch) (https://github.com/foyer-rs/foyer/pull/1294)

tisonkun:
@pmcgleenon Sorry it's a few time ago so I just noticed that you're the issue author. Then the message is clear and simple:

Please send a PR and ping me as a reviewer :D

pmcgleenon:
@tisonkun  thanks, sure I'll create a PR for it and tag you so that you can take a look.  The interface is similar to the datasketches

tisonkun:
@pmcgleenon any update here? This issue comes to me today and I'd like to review any PR that initializes the REQSketch impl.

pmcgleenon:
[...]

Issue #203, opened 2026-08-17
missing common_random support in library


jeffoodchain:
According to this: https://github.com/apache/datasketches-rust/pull/62#discussion_r2701211297, there's no random support in rust std.

I am currently writing `req` but I would need things like this, so I think it might be great to have one in our lib.

```rust
use std::cell::Cell;
use std::time::SystemTime;
use std::time::UNIX_EPOCH;

fn seed() -> u64 {
    let nanos = SystemTime::now()
        .duration_since(UNIX_EPOCH)
        .unwrap_or_default()
        .as_nanos() as u64;
    // xorshift64 degenerates to all zeros if seeded with zero.
    if nanos == 0 {
        0x2545_F491_4F6C_DD1D
    } else {
        nanos
    }
}

// Returns a pseudo-random bit.
pub(crate) fn random_bit() -> bool {
    thread_local! {
        static RNG_STATE: Cell<u64> = Cell::new(seed());
    }
    RNG_STATE.with(|state| {
        let mut x = state.get();
        x ^= x << 13;
        x ^= x >> 7;
        x ^= x << 17;
        state.set(x);
        (x & 1) == 1
    })
}

```

tisonkun:
Anywhere we can integrate with these methods? I don't remember we need random_common right now or anywhere.

You may add one when there is a real use case instead of adding utilities beforehand.

jeffoodchain:
noted. :), then I'll try to add one in req I'm working on.

tisonkun:
Now that I can see we use `rand` in KLL and REQ sketches, while I think `rand` is the de facto standard to use, it's fair enough for datasketches to use a more focused solution.

jeffoodchain:
I think it's not worthy to use `rand` for now, maybe we should switch to https://github.com/smol-rs/fastrand. `rand`'s introducing too many dependency and now I think `fastrand` is enough for our needs.

tisonkun:
.. or a vendored (optimized) version narrowed into our use case.

jeffoodchain:
I see. fair enough

PR #204, opened 2026-08-18
feat(req): port of reqsketch-rs 

pmcgleenon:
This is a port of [reqsketch-rs](https://github.com/pmcgleenon/reqsketch-rs) as discussed here https://github.com/apache/datasketches-rust/issues/90.   This implementation is influenced heavily by the apache datasketches C++ implementation.   

Some things present in reqsketch-rs that are not part of this PR:
 - [benchmarks](https://github.com/pmcgleenon/reqsketch-rs/blob/main/benches/core_benchmarks.rs), I left these out since I didn't see any other benchmarks.  Maybe we can add some in a later PR for all the sketches
 - program build the plots [rank error plots](https://github.com/pmcgleenon/reqsketch-rs/blob/main/examples/req_rank_error.rs).  There are a bunch of dependencies needed for this so I left it out.  But I think it's a great visualization of the reqsketch algorithm and helps with validation of the results since the apache-datasketches project [has similar plots](https://datasketches.apache.org/docs/REQ/ReqSketch.html) and they can be easily compared to validate the results

@tisonkun  FYI if you get a chance please take a look


tisonkun (commented):
Thanks for your contribution @pmcgleenon! I'll review this patch in this week.

datasketches/src/codec/family.rs
```diff
@@ -80,6 +80,15 @@ impl Family {
         max_pre_longs: 5,
     };
 
+    /// Relative Error Quantiles (REQ) sketch.
+    #[cfg(feature = "req")]
+    pub const REQ: Family = Family {
+        id: 17,
+        name: "REQ",
+        min_pre_longs: 2,
+        max_pre_longs: 4,
```
tisonkun: From datasketches-java, these fields are:

```java
REQ(17, "REQ", 1 /* minPreLongs */, 2 /* maxPreLongs */),
```

But we don't use these fields for REQSketch anyway.

datasketches/tests/serde_tests/req.rs
tisonkun, line 1: This should put under `serde_test` and need not to repeat `serialization_test_data` helper.

datasketches/Cargo.toml
```diff
@@ -87,9 +93,14 @@ name = "tuple_test"
 path = "tests/tuple_test/main.rs"
 required-features = ["tuple"]
 
+[dependencies]
+rand = { workspace = true, optional = true }
```
tisonkun, line 103: This can be relevant to https://github.com/apache/datasketches-rust/issues/203.

cc @jeffoodchain

We may use `rand` as a starting point and conclude before the next release.

datasketches/Cargo.toml
```diff
@@ -87,9 +93,14 @@ name = "tuple_test"
...
+
 [dev-dependencies]
+approx = { workspace = true }
```
tisonkun, line 107: `googletest` provides:

* https://docs.rs/googletest/latest/googletest/matchers/fn.near.html
* https://docs.rs/googletest/latest/googletest/matchers/fn.approx_eq.html

.. and we may need one more test lib. But you can leave it to me to do a global alignment.

datasketches/src/req/sketch.rs
```diff
@@ -0,0 +1,833 @@
...
+    /// Returns the largest item ever observed, or `None` if empty.
+    pub fn max_item(&self) -> Option<&T> {
+        self.max_item.as_ref()
+    }
+
+    /// Updates the sketch with a new item.
+    ///
+    /// NaN inputs are silently ignored for floating-point types, matching the behavior
+    /// of the Java reference implementation (`checkNaNUpdate`). This is intentional and
+    /// documented in the cross-language differences doc.
```
tisonkun: > in the cross-language differences doc.

Seems not included in this PR. We may rewrite this doc comment a bit.

datasketches/src/req/sketch.rs
```diff
@@ -0,0 +1,833 @@
...
+                "k must be in [{}, {}], got {k}",
+                MIN_K, MAX_K
+            )));
+        }
+        if k % 2 != 0 {
+            return Err(Error::invalid_argument(format!("k must be even, got {k}")));
+        }
+        self.k = k;
+        Ok(self)
+    }
```
tisonkun, line 803: Other builders use a `panic` flavor on setters.

I may prefer to follow the same flavor and propose a new flavor if desired later.

Besides, once we check `k` here, we may not need to call `ReqSketch::try_new` on build which check the same conditions.

And we can have a `ReqSketch::new` as the panic version for `try_new`. The current no-param `new` can be inlined to `default` and only `default`.

datasketches/src/req/sketch.rs
```diff
@@ -0,0 +1,833 @@
...
+
+    /// Sets the rank accuracy.
+    pub fn rank_accuracy(mut self, ra: RankAccuracy) -> Self {
+        self.rank_accuracy = ra;
+        self
+    }
```
tisonkun:
```suggestion
    pub fn rank_accuracy(mut self, rank_accuracy: RankAccuracy) -> Self {
        self.rank_accuracy = rank_accuracy;
        self
    }
```

nit: public API would prefer explicit over abbr.

datasketches/src/req/sketch.rs
```diff
@@ -0,0 +1,833 @@
...
+        writeln!(f, "  num retained     : {}", self.num_retained)?;
+        writeln!(f, "  num levels       : {}", self.compactors.len())?;
+        writeln!(f, "  estimation mode  : {}", self.is_estimation_mode())?;
+        if let (Some(min), Some(max)) = (&self.min_item, &self.max_item) {
+            writeln!(f, "  min item         : {min}")?;
+            writeln!(f, "  max item         : {max}")?;
+        }
+        Ok(())
+    }
+}
```
tisonkun, line 832: Good point. I wonder if we'd have a tracking issue to `impl fmt::Display` for all sketches when applicible. This is what datasketches-java does IMO, but the format can be considered once more.

cc @notfilippo @ariesdevil @ZENOTME

datasketches/src/req/union.rs
```diff
@@ -0,0 +1,95 @@
...
+
+use super::RankAccuracy;
+use super::sketch::ReqSketch;
+use super::value::ReqValue;
+use crate::error::Error;
+
+/// Combines multiple [`ReqSketch`]es into a single result.
+///
+/// Internally wraps a `ReqSketch` configured for union semantics. The C++
+/// equivalent is `req_union<T>`.
```
tisonkun: I see no `req_union<T>` in datasketches-cpp.

datasketches/src/req/union.rs
```diff
@@ -0,0 +1,95 @@
...
+    /// Creates a new union with the given `k` and rank accuracy.
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if `k` is invalid (see [`ReqSketch::try_new`]).
+    pub fn try_new(k: u16, rank_accuracy: RankAccuracy) -> Result<Self, Error> {
+        Ok(Self {
+            inner: ReqSketch::try_new(k, rank_accuracy)?,
+        })
+    }
```
tisonkun, line 50: Similarly, I'd prefer `new` to be a panic version of `try_new` and leave the no-param `new` as `default`.

datasketches/src/req/union.rs
tisonkun: cc @AlexanderSaydakov @leerho You can review the `ReqUnion` design here and give some high-level design comments.

tisonkun:
Glanced at the surface and commented above. Below is a review comment from Codex. I'll dive into implementation details later this week. And I think it's OK to merge a starting implementation and iterate later. But let's discuss existing items first to decide when to converge.

[CODEX COMMENT STARTS]
I found two correctness issues that I believe need to be fixed before this can be merged.

1. **`Compactor::merge` breaks the ordering invariant required by the wire format**

In `datasketches/src/req/compactor.rs`, `Compactor::merge` appends the other compactor’s items and marks the result as unsorted. This is safe only while the sketch remains in memory, because `rank()` falls back to a linear scan.

However, the REQ wire format stores a sorted flag only for level 0. Higher levels are implicitly required to be sorted, and `ReqSketch::deserialize` consequently marks every level above 0 as sorted. A merged sketch can therefore serialize unsorted higher-level items, which are then binary-searched as if they were sorted after deserialization.

I reproduced this with two HRA sketches using `k = 12`, containing `1000..1072` and `0..72`. After merging the latter into the former and performing a serialize/deserialize round trip, the restored sketch returned different results from its sorted view for the same rank query. In one run:

```text
restored.rank(0.0, Inclusive)               = 0.097222...
restored.sorted_view().rank(0.0, Inclusive) = 0.013888...
```

This also means a merged image produced by Rust may be interpreted incorrectly by Java or C++.

Both reference implementations preserve ordering during compactor merge by sorting and performing an ordered merge. I suggest doing the same here, including validating that the two compactors have the same `lg_weight`.

Please also add a regression test that:

- builds two estimation-mode sketches with reversed/disjoint value ranges;
- merges them;
- serializes and deserializes the result; and
- verifies that direct rank queries remain equal to sorted-view rank queries.

2. **Deserialization accepts impossible compactor states that later panic or violate public API contracts**

The current validation checks `section_size_raw` and requires `lg_weight < 64`, but it does not validate several structural invariants controlled by the serialized input.

Two concrete examples:

- An image with `num_sections = 0` is accepted. Calling `merge()` on the resulting sketch then panics in `ensure_enough_sections()` when evaluating `num_sections - 1`.
- A single-level image with `lg_weight = 63` and `n = 1` is accepted. Its inclusive rank query returns approximately `9.22e18`, despite `rank()` promising a result in `[0, 1]`.

A malformed serialized image should not be able to create a sketch for which safe public operations panic or return impossible results.

Before constructing the sketch, I suggest validating at least:

- each compactor’s `lg_weight` equals its level index;
- `num_sections`, `section_size_raw`, and `state` form a valid, non-zero compactor configuration;
- the number of levels is representable by the `u64` weight model;
- `sum(num_items * 2^level) == n`;
- levels declared as sorted are actually sorted; and
- extrema and retained values are consistent and contain no invalid NaN state.

The relevant arithmetic should also use checked operations. Since these failures originate from malformed serialized input, they should return `ErrorKind::InvalidData`, not `InvalidArgument`.

Once these two issues are addressed, the existing algorithm and cross-language compatibility coverage will be on much firmer ground.
[CODEX COMMENT END]

datasketches/src/req/compactor.rs
```diff
@@ -0,0 +1,604 @@
...
+
+    /// Returns the weight (2^lg_weight) for items in this compactor.
+    pub(super) fn weight(&self) -> u64 {
+        1u64 << self.lg_weight
+    }
+
+    // Private helper methods
+
+    fn ensure_enough_sections(&mut self) -> bool {
+        let ssr = self.section_size_raw / (2.0_f32).sqrt();
```
jeffoodchain: misc: `std::f32::consts::SQRT_2` might be a great idea here.

datasketches/src/req/compactor.rs
```diff
@@ -0,0 +1,604 @@
...
+        let ne = nearest_even(ssr);
+
+        const MIN_K: u32 = 4; // matches datasketches-cpp
```
jeffoodchain: I'm curious why not using the one declared in `req/mod.rs`

jeffoodchain (commented):
Just have some insight for `compactor.rs` since I tried to work on this component recently.

datasketches/Cargo.toml
jeffoodchain: thanks, I think it's better to have `rand` as dependency.

pmcgleenon: [fastrand](https://crates.io/crates/fastrand) might also be a good option.   I saw some benchmarks [here](https://github.com/hoxxep/rapidrand#performance)

datasketches/src/codec/family.rs
pmcgleenon: Have updated it here https://github.com/apache/datasketches-rust/pull/204/changes/d5f9d2f08890160b4372f62ca0fa74d8c27e3296

datasketches/src/req/sketch.rs
pmcgleenon: Have removed the reference here https://github.com/apache/datasketches-rust/pull/204/changes/d5f9d2f08890160b4372f62ca0fa74d8c27e3296

datasketches/src/req/sketch.rs
pmcgleenon: I prefer to avoid run-time crashes for these kind of errors and was taking a similar approach to the C++ error handling which is using exceptions rather than asserts.     I'm happy to align with the rest of the sketches here if needed.   I think nopanic for this crate would be an improved developer experience, maybe we can consider that for the future.

datasketches/src/req/sketch.rs
pmcgleenon: Have fixed this here https://github.com/apache/datasketches-rust/pull/204/changes/d5f9d2f08890160b4372f62ca0fa74d8c27e3296

datasketches/src/req/union.rs
pmcgleenon: I removed the reference to req_union here https://github.com/apache/datasketches-rust/pull/204/changes/d5f9d2f08890160b4372f62ca0fa74d8c27e3296

We could collapse it back into ReqSketch or leave it

datasketches/src/req/compactor.rs
pmcgleenon: thanks, I've used it here https://github.com/apache/datasketches-rust/pull/204/changes/d5f9d2f08890160b4372f62ca0fa74d8c27e3296

datasketches/src/req/compactor.rs
pmcgleenon: Good catch, updated https://github.com/apache/datasketches-rust/pull/204/changes/d5f9d2f08890160b4372f62ca0fa74d8c27e3296

pmcgleenon:
I can create a couple of stacked PRs for these two?   

> [CODEX COMMENT STARTS] I found two correctness issues that I believe need to be fixed before this can be merged.
> 
> 1. **`Compactor::merge` breaks the ordering invariant required by the wire format**

I had been looking at introducing ordering in merge in a different context and have something working there.  This looks like a good catch

> 
> 2. **Deserialization accepts impossible compactor states that later panic or violate public API contracts**

Seems reasonable, I'll take a look at this one

datasketches/tests/serde_tests/req.rs
pmcgleenon: moved it here https://github.com/apache/datasketches-rust/pull/204/changes/d5f9d2f08890160b4372f62ca0fa74d8c27e3296

tisonkun:
> I can create a couple of stacked PRs for these two?

For sure. I'm kinda a Git guru so any shape that has a good final status should be fine :D

datasketches/Cargo.toml
tisonkun: If we do need external deps here, let's start with `rand`.

datasketches/src/req/sketch.rs
tisonkun: Now tracked in https://github.com/apache/datasketches-rust/issues/212.

tisonkun:
> * [benchmarks](https://github.com/pmcgleenon/reqsketch-rs/blob/main/benches/core_benchmarks.rs), I left these out since I didn't see any other benchmarks.  Maybe we can add some in a later PR for all the sketches

Now we have one under `datasketches/benches`. If the bench-only dependencies get complicated, we can have a standalone local crate like [apache/asyncband's benchmarks](https://github.com/apache/asyncband/tree/879dc4eec566bc2f210d1a8b924fa78ce0f53189/benchmarks).

But we can introduce the benchmarks in a follow-up PR.

tisonkun:
Now I'm trying to resolve the major comments on this PR today or tomorrow and merge it so @jeffoodchain or anyone else can work based on a basic REQ impl.

We're still in an early development stage so there is no reason to pending a PR for a long time :D

datasketches/src/req/sketch.rs
Xuanwo: Maybe we can have a build() API that consumes the Builder and returns the final result.

tisonkun: Reasonable. I'm trying to find a way to merge this PR without too many extended requirements and do a thorough refactor of all related API later.

Xuanwo: I can help with that.

datasketches/src/req/union.rs
```diff
@@ -0,0 +1,94 @@
...
+//! REQ union — combines REQ sketches into a single result.
+
+use super::RankAccuracy;
+use super::sketch::ReqSketch;
+use super::value::ReqValue;
+use crate::error::Error;
+
+/// Combines multiple [`ReqSketch`]es into a single result.
+///
+/// Internally wraps a `ReqSketch` configured for union semantics.
```
tisonkun:
```suggestion
/// Internally wraps a `ReqSketch` configured for union semantics.
```

datasketches/src/req/sketch.rs
tisonkun: OK. Then let me merge this PR as is and I'll push my follow-ups in a separate PR as well.

tisonkun (approved):
Thanks for your contribution @pmcgleenon!

PR #208, opened 2026-08-23
feat: add xor filter

tisonkun:
## Summary

- add immutable `XorFilter` construction and lookup with 8- and 16-bit fingerprints behind the `xor` feature
- support both ordinary Rust `Hash` values and precomputed 64-bit hashes, with duplicate removal and bounded seeded construction retries
- add family 22 serialization, defensive deserialization, public API documentation, README/changelog entries, and dedicated behavior/serialization tests
- add construction and query benchmarks next to the existing sketch benchmarks, with the `xorf` crate as a reference implementation for comparison

## Design

The implementation follows the three-partite peeling construction from Graf and Lemire's xor-filter paper directly, without adding a third-party runtime dependency. `XorFilterBuilder::build` returns `Result` because construction is deliberately bounded at 100 seed attempts instead of risking an unbounded loop. The default base construction seed is deterministic, and callers can provide another seed when needed.

Serialization follows the 3-long family 22 layout proposed in apache/datasketches-java#740. Tests use images generated by that Java implementation and verify that Rust produces byte-for-byte identical images for the same values and explicit seed. The initial API intentionally exposes only the 8- and 16-bit variants defined by that portable format.

Deserialization validates the family, version, fingerprint width, hash count, signed lengths, indexable capacity, item count, and payload length before constructing a filter.

## Testing

- `cargo x check`
- `cargo x test`
- `cargo x lint`
- `cargo bench --package benchmarks -- xor` (comparative baseline against `xorf`)
- byte-for-byte compatibility against 8- and 16-bit images generated from apache/datasketches-java#740


tisonkun:
FYI @leerho @freakyzoidberg I'm trying to implement XorFilter in Rust as well

PR #211, opened 2026-08-24
feat: return Option from intersection to_sketch before first update

### LINKED ISSUE #192: Define intersection result extraction before the first update
ThetaIntersection::to_sketch and TupleIntersection::to_sketch currently panic when called before the first update. Both operators expose has_result, and tests explicitly cover the panic, but absence of an intersection state is part of the public state machine and deserves a deliberate API contract.

We should decide whether result extraction should:

- return Option when no result is a normal state;
- return Result when calling before update is considered recoverable misuse; or
- retain the panic, with a clear invariant and local Panics documentation.

Points to settle:

- Apply the same contract to Theta and Tuple intersections.
- Decide whether has_result remains useful after result extraction becomes typed.
- Define the mathematical meaning of zero updates; the identity for intersection cannot be represented as a finite sketch.
- Consider reset if it is added to the operators later.
- Check Java and C++ precedent, while keeping the Rust API misuse-resistant.
- Document migration if the return type changes before 1.0.

This issue intentionally records the design question rather than prescribing Option or Result immediately.

VedantMadane:
## Summary

feat: return Option from intersection to_sketch before first update

## Changes

- ThetaIntersection::to_sketch and TupleIntersection::to_sketch return Option
- tests: expect None instead of panic before first update

Fixes #192

### DIFF
--- a/datasketches/src/thetafamily/theta/intersection.rs
+++ b/datasketches/src/thetafamily/theta/intersection.rs
@@ -25,8 +25,9 @@ use crate::thetacommon::intersection::IntersectionState;
 
 /// Stateful intersection operator for Theta sketches.
 ///
-/// Before the first [`update`](Self::update), the result is undefined; use
-/// [`has_result`](Self::has_result) to check.
+/// A newly created operator has no result. [`has_result`](Self::has_result) returns `false` and
+/// [`to_sketch`](Self::to_sketch) returns `None` until the first successful
+/// [`update`](Self::update).
 #[derive(Debug)]
 pub struct ThetaIntersection {
     state: IntersectionState<ThetaEntry, NoopIntersectionPolicy>,
@@ -63,7 +64,7 @@ impl ThetaIntersection {
         self.state.update(sketch)
     }
 
-    /// Returns whether this operator has received at least one update.
+    /// Returns `true` after the first successful [`update`](Self::update).
     pub fn has_result(&self) -> bool {
         self.state.has_result()
     }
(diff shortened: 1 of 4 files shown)


datasketches/src/thetafamily/tuple/intersection.rs
```diff
@@ -79,7 +79,7 @@ use crate::tuple::sketch::TupleSketchView;
 /// intersection.update(&a).unwrap();
 /// intersection.update(&b).unwrap();
 ///
-/// let result = intersection.to_sketch(true);
+/// let result = intersection.to_sketch(true).expect("after update");
```
tisonkun:
```suggestion
/// let result = intersection.to_sketch(true).unwrap();
```

tisonkun (approved):
Thanks for your contribution @VedantMadane! LGTM.

cc @ZENOTME @ariesdevil

datasketches/tests/tuple_test/intersection.rs
```diff
@@ -49,14 +49,13 @@ fn has_result_tracks_the_first_update() {
...
+    assert_eq!(intersection.to_sketch(true).unwrap().num_retained(), 10);
 }
 
 #[test]
-#[should_panic(expected = "before first update")]
-fn result_before_first_update_panics() {
-    let intersection = TupleIntersection::new(SumPolicy);
-    let _ = intersection.to_sketch(true);
+fn result_before_first_update_returns_none() {
+    let intersection = TupleIntersection::<IdentityPolicy>::new();
```
tisonkun:
```suggestion
    let intersection = TupleIntersection::new(SumPolicy);
```

LLM Hallucination?

datasketches/src/thetafamily/theta/intersection.rs
```diff
@@ -75,16 +75,15 @@ impl ThetaIntersection {
...
-    ///
-    /// Panics if called before the first [`update`](Self::update).
-    pub fn to_sketch(&self, ordered: bool) -> CompactThetaSketch {
-        assert!(
-            self.state.has_result(),
-            "ThetaIntersection::to_sketch() called before first update()"
-        );
+    /// Returns `None` if called before the first [`update`](Self::update).
+    /// Absence of a result is part of the public state machine; use
+    /// [`has_result`](Self::has_result) or this `Option` return instead of panicking (#192).
```
tisonkun: We should improve this doc; at least we should not suddenly tell `#192` or similar thing that is too detailed.

tisonkun: Addressed in d1aaa26. I rewrote the Theta and Tuple docs around the observable contract: a new intersection returns `None`; after a successful update it returns `Some`, including when the intersection is empty. I also removed the issue reference and aligned the stale Theta test name.

PR #215, opened 2026-08-25
chore(deps): upgrade rand to 0.10 and use quickcheck

tisonkun:
## Summary

- upgrade the workspace `rand` requirement from 0.9.2 to 0.10.0; the lockfile resolves the latest compatible release, 0.10.2
- replace proptest with QuickCheck 1.1 for the existing REQ property tests, keeping the test graph on rand 0.10
- test the generic rank and ordering invariants directly with supported integer domains instead of manufacturing bounded floats from integer inputs
- remove redundant per-module `req` feature gates because the `req_test` target is already gated by `required-features = ["req"]`

## Property-testing choice

The REQ properties in this repository exercise generic ordering, rank consistency, bounds, and monotonicity. They do not require floating-point-specific behavior, so the QuickCheck properties now use native `u64`, `i64`, and `u32` inputs that QuickCheck can generate and shrink directly. Existing focused tests continue to cover floating-point behavior such as NaN handling and total ordering.

The explicit `QuickCheck` builder keeps each collection-length domain visible and requires 256 valid cases. This avoids custom `Arbitrary` wrappers and avoids treating QuickCheck's arbitrary integers as if they were a uniform bounded-float generator.

QuickCheck 1.1 and datasketches now share rand 0.10.2, and the proptest-only dependency subtree is removed. The REQ compactor coin flip remains unchanged because `rand::random` is source-compatible with rand 0.10.

No changelog entry is included because this is dependency and test maintenance without a supported or observable behavior change.

## Validation

- `cargo test --locked -p datasketches --features req --test req_test property::`
- `cargo x check`
- `cargo x test`
- `cargo x lint`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -170,6 +170,17 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "9330f8b2ff13f34540b44e946ef35111825727b38d33286ef986142615121801"
 
 [[package]]
+name = "chacha20"
+version = "0.10.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "d524456ba66e72eb8b115ff89e01e497f8e6d11d78b70b1aa13c0fbd97540a81"
+dependencies = [
+ "cfg-if",
+ "cpufeatures",
+ "rand_core 0.10.1",
+]
+
+[[package]]
 name = "clap"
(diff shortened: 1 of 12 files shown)


tisonkun:
I'm not fully confident to `quickcheck` but since we bump to rand 0.10 let's use it for now.

PR #220, opened 2026-08-25
fix: compactor merge

### LINKED ISSUE #219: REQ merge must preserve compactor ordering across serialization round trips
## Description

`Compactor::merge` appends the other compactor's items and marks the result as unsorted. This is safe only while the in-memory `is_sorted` flag is preserved. The REQ wire format stores that flag for level 0 only, and `ReqSketch::deserialize` assumes every higher level is sorted.

As a result, serializing and deserializing a merged sketch can silently change rank results: higher-level items remain physically unsorted, but the restored sketch binary-searches them as if they were sorted. Images produced by Rust may also be interpreted incorrectly by the Java and C++ implementations, whose merge paths preserve sorted higher levels.

Relevant code:

- [`Compactor::merge` appends items and sets `is_sorted = false`](https://github.com/apache/datasketches-rust/blob/bcb83e629ae86eb8619bb7eb0a372a172c94b79c/datasketches/src/req/compactor.rs#L119-L129)
- [`ReqSketch::deserialize` marks all levels above level 0 as sorted](https://github.com/apache/datasketches-rust/blob/bcb83e629ae86eb8619bb7eb0a372a172c94b79c/datasketches/src/req/sketch.rs#L672-L676)

## Reproduction

```rust
use datasketches::req::{ReqSketch, SearchCriteria};

let mut high = ReqSketch::<f64>::new();
let mut low = ReqSketch::<f64>::new();

for value in 1000..=1072 {
    high.update(value as f64);
}
for value in 0..=72 {
    low.update(value as f64);
}

high.merge(&low).unwrap();
let restored = ReqSketch::<f64>::deserialize(&high.serialize()).unwrap();
let view = restored.sorted_view();

for value in 0..=1072 {
    let value = value as f64;
    assert_eq!(
        restored.rank(&value, SearchCriteria::Inclusive).unwrap(),
        view.rank(&value, SearchCriteria::Inclusive).unwrap(),
    );
}
```

On current `main`, one run reported `0.0958904109589041` from `ReqSketch::rank` and `0.0136986301369863` from `SortedView::rank` for the same query after the round trip.

## Expected fix

- Make `Compactor::merge` preserve the ordering invariant required by the wire format. Sorting unsorted inputs and performing an ordered merge, as the Java and C++ implementations do, is one possible approach.
- Check or assert that the two compactors have the same `lg_weight` before merging them.
- Add a regression test that merges sketches containing reverse, disjoint ranges, serializes and deserializes the result, and verifies that direct ranks equal sorted-view ranks across representative queries.
- Keep serialization-related coverage in `datasketches/tests/serde_tests/req.rs`, following `CONTRIBUTING.md`.

Introduced with the REQ implementation in #204.

jeffoodchain:
closes #219

### DIFF
--- a/datasketches/src/req/compactor.rs
+++ b/datasketches/src/req/compactor.rs
@@ -118,10 +118,14 @@ where
 
     /// Merges items from another compactor into this one.
     pub(super) fn merge(&mut self, other: &Self) {
+        debug_assert_eq!(self.lg_weight, other.lg_weight);
         self.state |= other.state;
-        self.items.extend_from_slice(&other.items);
         if !other.items.is_empty() {
-            self.is_sorted = false;
+            // make sure both items are sorted.
+            self.sort();
+            let mut other_items = other.items.clone();
+            other_items.sort_unstable_by(|a, b| a.total_cmp(b));
+            self.merge_sorted(&other_items);
         }
         // OR-ing the schedule counters can advance state past several doubling
         // thresholds at once. Loop until no more doublings are needed (C++:
--- a/datasketches/tests/serde_tests/req.rs
+++ b/datasketches/tests/serde_tests/req.rs
@@ -231,6 +231,31 @@ fn deserialize_truncated_raw_items() {
     assert_that!(result, err(anything()));
 }
 
+#[test]
+fn merge_preserves_order_across_serde_round_trip() {
+    let mut high = ReqSketch::<f64>::new();
+    let mut low = ReqSketch::<f64>::new();
+
+    for value in 1000..=1072 {
+        high.update(value as f64);
+    }
+    for value in 0..=72 {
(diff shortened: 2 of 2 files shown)


tisonkun (approved):
Thanks for your contribution!

tisonkun:
@pmcgleenon 

I remembered you wrote:

> I had been looking at introducing ordering in merge in a different context and have something working there. This looks like a good catch

Not sure if this PR and #222 resolves all the issues in your mind. Or if there can be any other issues, feel free to track them down first or directly submit new patches :D

pmcgleenon:
I think this covers it, the only point was the additional clone which you have already removed.  Nice work @jeffoodchain !
The merge operation is heavyweight and could benefit from further optimization

PR #221, opened 2026-08-25
fix: reject invalid REQ compactor states on deserialize

### LINKED ISSUE #218: Reject invalid REQ compactor states during deserialization
## Description

REQ deserialization currently accepts compactor states that cannot be produced by a valid sketch. Those states can later panic in safe public APIs or violate documented return-value contracts.

Two concrete examples on current `main`:

1. A single-level image with `num_sections = 0` deserializes successfully. A subsequent `ReqSketch::merge` panics at `1u64 << (self.num_sections - 1)` because the subtraction underflows.
2. A single-level image whose level-0 compactor declares `lg_weight = 63` and contains one item deserializes successfully. Its inclusive rank for that item is approximately `9.22e18`, even though `ReqSketch::rank` promises a result in `[0.0, 1.0]`.

Relevant code:

- [`Compactor::deserialize` validates only broad `section_size_raw` and `lg_weight` bounds](https://github.com/apache/datasketches-rust/blob/bcb83e629ae86eb8619bb7eb0a372a172c94b79c/datasketches/src/req/compactor.rs#L355-L416)
- [`ensure_enough_sections` assumes `num_sections` is nonzero](https://github.com/apache/datasketches-rust/blob/bcb83e629ae86eb8619bb7eb0a372a172c94b79c/datasketches/src/req/compactor.rs#L287-L300)
- [`ReqSketch::deserialize` accepts the reconstructed compactors without validating cross-field invariants](https://github.com/apache/datasketches-rust/blob/bcb83e629ae86eb8619bb7eb0a372a172c94b79c/datasketches/src/req/sketch.rs#L661-L714)

## Expected fix

Validate wire-controlled compactor state before constructing a `ReqSketch`. At minimum:

- Reject `num_sections = 0` and validate it against states the REQ section schedule can produce.
- Require each compactor's `lg_weight` to equal its enclosing level index.
- Use checked arithmetic to calculate the retained weighted count and require it to equal the serialized stream length `n`.
- Return `ErrorKind::InvalidData` for malformed serialized state instead of `InvalidArgument`.
- Ensure malformed input is rejected by `deserialize` rather than causing a later public operation to panic or return a rank outside `[0.0, 1.0]`.

Please add focused malformed-image regression tests to `datasketches/tests/serde_tests/req.rs`, including the two cases above. The existing `single_level_image` helper can be reused.

Introduced with the REQ implementation in #204.

cestercian:
Fixes #218

## Summary

REQ deserialization accepted compactor states that a valid sketch cannot produce. Those images later panicked in `ReqSketch::merge` or returned a `rank()` outside `[0.0, 1.0]`. Deserialize now rejects those states with `ErrorKind::InvalidData`.

## Changes Made

- `Compactor::deserialize` rejects `num_sections` values that the REQ doubling schedule cannot produce, including `num_sections = 0`. Malformed wire fields now return `ErrorKind::InvalidData` instead of `InvalidArgument`.
- `ReqSketch::deserialize` requires each compactor's `lg_weight` to equal its enclosing level index.
- `ReqSketch::deserialize` uses checked arithmetic for the retained weighted count and requires it to equal serialized `n`.
- Regression tests in `datasketches/tests/serde_tests/req.rs` cover the two malformed single-level images from the issue.

## Related Issue

https://github.com/apache/datasketches-rust/issues/218

## Testing Steps

1. `cargo x test` (from the repo root; run `cargo x prepare-testdata` first if serialization fixtures are missing)
2. `cargo x lint`
3. Confirm the new serde tests fail the two crafted images with `ErrorKind::InvalidData`:
   - single-level image with `num_sections = 0`
   - single-level image with level-0 `lg_weight = 63` and one item


datasketches/src/req/sketch.rs
```diff
@@ -716,6 +729,19 @@ impl<T: ReqValue> ReqSketch<T> {
 
     // --- Internal ---
 
+    fn retained_weighted_count(compactors: &[Compactor<T>]) -> Result<u64, Error> {
```
jeffoodchain: I think maybe no need to extract this as a function since it is only used once.

datasketches/src/req/sketch.rs
jeffoodchain: I think there are some `Error::invalid_argument` needed to change to `Error::deserial` in this function.

jeffoodchain (commented):
just a few comments from my opinion 
- `retained_weighted_count` is only used once, maybe we could leave it without extracting it as an internal function
- there are some `Error::invalid_argument` in the `deserialize` function, should be changed to `Error::deserial` as well, according to the issue description.

tisonkun (changes requested):
There is one remaining correctness issue: a crafted image with `num_retained >= max_nom_size` still passes the new checks. Since `update()` compresses only when the two values are equal, the first update moves past the threshold and the sketch never compacts again.

Minimal reproducer (the level-0 nominal capacity is `2 * 12 * 3 = 72`):

```rust
let mut bytes = vec![2, 1, 17, 8, 12, 0, 1, 0];
bytes.extend_from_slice(&0u64.to_le_bytes()); // state
bytes.extend_from_slice(&12.0f32.to_le_bytes());
bytes.extend_from_slice(&[0, 3, 0, 0]); // lg_weight, num_sections, padding
bytes.extend_from_slice(&72u32.to_le_bytes());
for i in 0..72 {
    bytes.extend_from_slice(&(i as f32).to_le_bytes());
}
let mut sketch = ReqSketch::<f32>::deserialize(&bytes).unwrap();
assert_eq!(sketch.num_retained(), sketch.total_nominal_capacity());
sketch.update(72.0);
assert_eq!(sketch.num_retained(), 73); // capacity remains 72; future updates never compact
```

Please reject `retained_count >= total_nominal_capacity` during deserialization using checked arithmetic, and add `==`/`>` boundary regression coverage. Changing `update()` to `>=` would also be useful as defense in depth, but should not replace rejecting the malformed image.

The new changelog bullet should also be removed. REQ was introduced after `0.4.0` and is still entirely unreleased, so this is pre-release hardening of the new feature; the release-to-release observable change is already covered by the existing REQ entry under New features.

tisonkun (changes requested):
Thanks — the previous capacity issue and changelog point are fixed. I still found one correctness blocker on `dbeab56`.

`deserialize` trusts `IS_LEVEL_ZERO_SORTED`, but `rank()` then uses `partition_point`, which requires the items to actually be sorted. This image is accepted, yet the two rank paths disagree:

```rust
let mut bytes = vec![2u8, 1, 17, 8 | 32, 12, 0, 1, 0]; // HRA | LEVEL_ZERO_SORTED
bytes.extend_from_slice(&0u64.to_le_bytes());
bytes.extend_from_slice(&12.0f32.to_le_bytes());
bytes.extend_from_slice(&[0, 3, 0, 0]); // lg_weight, num_sections, padding
bytes.extend_from_slice(&5u32.to_le_bytes());
for item in [3.0f32, 4.0, 5.0, 1.0, 2.0] {
    bytes.extend_from_slice(&item.to_le_bytes());
}
let sketch = ReqSketch::<f32>::deserialize(&bytes).unwrap();
assert_eq!(sketch.rank(&2.0, SearchCriteria::Inclusive).unwrap(), 0.0);
assert_eq!(sketch.sorted_view().rank(&2.0, SearchCriteria::Inclusive).unwrap(), 0.4);
```

Please validate the ordering whenever a compactor is declared sorted, including the raw-items path, or normalize the malformed input before storing `is_sorted = true`.

There is also a broader invariant gap: `section_size_raw` is checked only against global `MAX_K`, not the enclosing sketch's `k`. A five-item image with `k = 4`, `section_size_raw = 1024`, and `num_sections = 3` is accepted with nominal capacity `6144` instead of `24`, allowing wire data to bypass the configured space bound. Please pass `k` into compactor validation and validate the reachable `(k, section_size_raw, num_sections)` combination rather than each field independently.

For the tests, the current one-item non-raw “valid baseline” is not canonical: serializers use `RAW_ITEMS` for one-level sketches with `n <= 4`. Using five items would make the malformed cases isolate only the intended field. Direct tests for weighted-count mismatch/overflow and a nonzero invalid schedule value would also cover the newly added branches.

Could you also share which model and prompt/verification workflow were used for this PR? The revisions have been closely fitted to individual counterexamples while repeatedly missing adjacent serialized-state invariants (capacity, ordering, and the `k`/section relationship); knowing the setup may help us give more targeted implementation guidance.

tisonkun:
Here is a suggested prompt for the coding agent. The key is to reason from the complete deserialized-state contract instead of patching each counterexample independently:

> Read `CONTRIBUTING.md` in full. Review the complete REQ construction, update, merge, serialize, and deserialize paths, plus the pinned C++/Java compatibility fixtures and reference implementations. Do not modify code yet.
>
> First, enumerate every invariant that holds for a `ReqSketch` produced through safe Rust APIs and serialization. For every wire-controlled field, identify all downstream operations that rely on it and the concrete failure caused by violating it. Cover at least: header/RAW_ITEMS consistency; level count and `lg_weight`; the reachable relationship among `k`, `state`, `section_size_raw`, and `num_sections`; actual ordering whenever `is_sorted` is true; NaN/value assumptions; weighted retained count versus `n`; retained count versus nominal capacity; min/max consistency; and arithmetic/allocation bounds.
>
> Then implement one coherent validation boundary that rejects malformed images with `ErrorKind::InvalidData` before constructing a usable sketch. Avoid validation rules based only on global numeric ranges when fields are derived from other serialized state. Preserve every valid C++/Java fixture and Rust round trip.
>
> Add focused regression tests that start from canonical valid images and mutate exactly one invariant. Include boundary and neighboring cases, not only the reported examples. In particular cover false sorted claims, `k`/section-schedule mismatches, nonzero invalid section counts, weighted-count mismatch and overflow, and retained count at/above capacity. Ensure each test asserts the intended validation branch and error kind.
>
> Keep the implementation small: share algorithm constants instead of duplicating them, use a test helper for repeated `InvalidData` assertions, and do not add a changelog entry because REQ is still unreleased. Run `cargo x prepare-testdata`, `cargo x check`, `cargo x test`, and `cargo x lint`, and report the invariant table, implementation choices, compatibility evidence, and command results.

cestercian:
@tisonkun thanks for the invariant-table prompt. Earlier revisions were follow-ups against individual counterexamples (the issue images, then `MIN_K` / capacity). This round used your prompt as written: enumerate the deserialized-state contract first, then one `InvalidData` boundary, with tests that start from a canonical five-item image and mutate one field.

Implementation was Cursor cloud agents on this contributor fork. Verification was `cargo x prepare-testdata`, `cargo x check`, `cargo x test`, and `cargo x lint` (`serde_tests` 101 passed, including the C++/Java REQ fixtures). Java `section_size_raw` can land 2 ULPs off the C++/Rust doubling-schedule replay, so that check keeps the reachable `(k, state, num_sections)` relation and tolerates that float noise so the pinned TCK images still load.

Happy to adjust the validation or the 2-ULP tolerance if you want it stricter.

tisonkun (approved):
The correctness issues are addressed, the implementation has been simplified, and the full local and cross-platform CI suites pass.

@cestercian, please review the current diff end to end, especially the serialized section-schedule validation, Java/C++ compatibility, and the boundary between required safety checks and over-validation. Please flag any concerns or alternatives you think we should consider.

cestercian (commented):
Reviewed `d622d68` end to end against the C++ `ensure_enough_sections` / Java `ensureEnoughSections` predicates, the pinned TCK fixtures, and issue #218.

**Section schedule.** `has_reachable_section_configuration` is the right model. C++ doubles with `section_size_raw_ / sqrtf(2)` when `nearest_even(ssr) >= MIN_K`. Java doubles with `(float)(sectionSizeFlt / Math.sqrt(2.0))` and an extra `sectionSize > MIN_K` on the *current* integer size. The two candidates in `compatible_next_section_sizes` match those predicates, including the Java guard, and the BFS over both at every doubling is what you want for a sketch that can move between implementations. Bit-exact `to_bits` comparison is stricter and clearer than the earlier 2-ULP window: it accepts actual C++/Java (and mixed) results, and `deserialize_rejects_capacity_changing_float_drift` still rejects 1-ULP noise that would change `nearest_even`. The stale-config check (`state` past the next threshold only if *both* producers would still double) also matches “at least one implementation would have stopped.”

**Safety vs over-validation.** I agree with what was kept versus dropped:

- Required for #218 / `rank()` / `update()`: `lg_weight` equals the level, `state != u64::MAX`, reachable `(k, state, section_size_raw, num_sections)`, actual order when a level is treated as sorted, no NaN items, weighted retained count equals `n`, RAW_ITEMS consistent with `n` and level count.
- Higher levels always validated as sorted is required even though the flag is only `IS_LEVEL_ZERO_SORTED`; `count_below` uses `partition_point` there.
- Dropping `retained >= nominal capacity` on deserialize looks right for this issue. Over-capacity does not produce the #218 failures, and `update()` already compresses on `>=`. C++ deserialize does not check this either.
- Min/max vs current items is correctly *not* required in estimation mode: extrema can have been compacted away. NaN and `min > max` are the useful checks.

**Alternatives worth considering, none blocking:**

1. Optionally reject `retained > capacity` (never produced by the algorithm) while still allowing `==` as the pre-compress state. Not needed for #218; `update()` `>=` is enough defense in depth.
2. A short comment on `compatible_next_section_sizes` citing the two reference predicates would make the Java `sectionSize > MIN_K` guard less mysterious.
3. Out of scope here: leftover trailing bytes after the last item are still ignored. Same as C++.

TCK `cpp_compatibility` / `java_compatibility` plus `deserialize_accepts_java_minimum_section_schedule` cover the compatibility surface I care about. No other concerns on this diff.

tisonkun:
@cestercian I found and fixed one additional malformed-state path after the previous review: two individually accepted high compactor states could OR to `u64::MAX` during merge and then overflow on compaction.

The updated validation requires each compactor's history to fit within the serialized stream length (`state * 2 * level_weight <= n`, using checked arithmetic). This follows because every level compaction consumes at least two items at that level's weight, and `a | b <= a + b` preserves the bound across merges. The tests include both a finite inconsistent state and the complementary-state merge reproducer.

I also ran a fresh defect-first review of the complete merge diff against the C++/Java section-growth logic and found no remaining actionable issues. Could you review the latest diff end to end and flag any concerns or alternatives?

cestercian (commented):
Looked at `d3ed3cc` against the merge overflow you described.

**Stream-length bound.** `state.checked_mul(2).checked_mul(weight) <= n` is the right replacement for the standalone `state == u64::MAX` check. `MAX * 2` overflows, so that image is still rejected. Any complementary pair that ORs to `MAX` has to cover bit 63, so at least one side fails `checked_mul(2)` on deserialize and never becomes a merge input. The finite case (`state=501` vs `n=1000`) is the one the old equality check would have missed.

Using 2 items per compaction is the conservative lower bound (actual REQ compactions drop a full section). Weaker bound means more images are accepted; that is the right direction for C++/Java interop, and it is still strong enough to keep `a | b` from saturating.

No other concerns on this delta. Previous notes on the dual-path section schedule and the safety/over-validation cut still stand.

PR #224, opened 2026-08-25
fix(frequencies): reject out-of-range map-size fields on deserialize

jaideeppyne:
`FrequentItemsSketch::deserialize` used the `lg_max_map_size` / `lg_cur_map_size` header bytes to size and left-shift the backing hash map without validating their range. A corrupt header with a large `lg_max_map_size` drives `1usize << lg_max` in `with_lg_map_sizes` past the width of `usize`, panicking with "attempt to shift left with overflow" in debug builds and requesting an oversized allocation in release builds — even on the empty-sketch path, before any payload is read.

This validates the map-size fields up front, mirroring the C++ reference `check_size` (`lg_cur >= LG_MIN_MAP_SIZE`, `lg_cur <= lg_max`) and adding an upper bound on `lg_max` (`LG_MAX_MAP_SIZE = 30`, matching CountMin's `MAX_TABLE_ENTRIES`) that Rust needs to keep the shift and allocation well-defined. Malformed input now returns `Error::deserial` instead of crashing.

Adds regression tests for out-of-range `lg_max_map_size` (empty and non-empty headers), undersized `lg_cur_map_size`, and a valid header. CHANGELOG updated.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ All significant changes to this project will be documented in this file.
 
 ### Bug fixes
 
+* `FrequentItemsSketch::deserialize` now rejects headers whose `lg_max_map_size` or `lg_cur_map_size` fields are out of range instead of panicking (debug builds) or requesting an oversized allocation (release builds) on corrupt input.
 * T-Digest compression now handles `k = u16::MAX` without overflowing the scale normalization input.
 * T-Digest deserialization now validates declared payload lengths before allocating. Updating a deserialized digest whose unmerged buffer already exceeds the compression threshold now compresses it instead of allowing the buffer to grow without bound.
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,7 +18,7 @@ All significant changes to this project will be documented in this file.
 
 ### Bug fixes
 
-* `FrequentItemsSketch::deserialize` now rejects headers whose `lg_max_map_size` or `lg_cur_map_size` fields are out of range instead of panicking (debug builds) or requesting an oversized allocation (release builds) on corrupt input. It also bounds the initial backing-map allocation to the payload: an empty header builds the map at the minimum size instead of honoring `lg_cur_map_size` (so an 8-byte empty header claiming `lg_cur = 30` no longer drives a `1 << 30`-slot allocation), and a non-empty header is rejected when `active_items` is inconsistent with `lg_cur_map_size` or with the remaining serialized bytes.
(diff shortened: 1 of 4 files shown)


tisonkun (commented):
Thanks for your contribution @jaideeppyne!

I'm curious how/where you noticed this issue. This information can help us to understand our users better.

jaideeppyne:
Good question — I found it by fuzzing the deserialize paths with truncated and bit-flipped headers (wrapped in `catch_unwind`) to check that every sketch returns `Err` on corrupt bytes rather than panicking. Frequencies was the one that slipped through: `lg_max_map_size` feeds straight into `1 << lg_max` in `with_lg_map_sizes` before any range check, so it overflows the shift on the empty-sketch path before any payload is read. Not from a production incident — just hardening the parse surface. FWIW the CPC deserialize path also panics on some malformed input; I can send a separate follow-up for that once this one lands.

tisonkun:
Could you please take a look at this round-trip case on a 64-bit target?

```rust
#[test]
fn round_trips_supported_large_max_map_size() {
    let sketch = FrequentItemsSketch::<i64>::new(1usize << 31);
    let bytes = sketch.serialize();
    FrequentItemsSketch::<i64>::deserialize(&bytes).unwrap();
}
```

`new(1 << 31)` currently succeeds and only allocates the small initial map, while serialization writes `lg_max_map_size = 31`. With this PR, deserialization then rejects the same bytes because of the new `LG_MAX_MAP_SIZE = 30` cap. I reproduced the test locally and it fails with `lg_max_map_size must be at most 30, got 31`.

Is this an intended new limit? If so, the constructor/API contract may need the same validation and the compatibility change should be documented; otherwise the deserialize-only cap appears to break a previously supported self round-trip.

tisonkun:
One more issue with the accepted boundary: `(lg_max_map_size, lg_cur_map_size) = (30, 30)` still passes the new validation and immediately attempts a very large allocation.

A minimal 8-byte empty header is:

```rust
let bytes = [1, 1, 10, 30, 30, 5, 0, 0];
FrequentItemsSketch::<i64>::deserialize(&bytes);
```

This reaches `ReversePurgeItemHashMap::new(1usize << 30)`. On a 64-bit target with `i64` items, the three backing arrays require roughly 26 GiB in total (`Option<i64>` + `u64` + `u16` per slot), so the process may abort or be killed rather than return `Error::deserial`.

I did not execute this exact case without a memory limit because it is intentionally OOM-inducing. Also, the test described as exercising the “exact upper bound” currently uses `lg_max = 10`, so it does not cover this path. Could you review whether accepting `lg_cur = 30` is really the intended resource limit?

jaideeppyne:
Good catch — you're right that `(30, 30)` slipped through and the empty path was still sizing the map straight from the header. I pushed a follow-up: the empty case now builds the map at `LG_MIN_MAP_SIZE` (matching the Java/C++ heapify, which ignores `lg_cur` when the sketch is empty), so `[1,1,10,30,30,5,0,0]` deserializes to an empty sketch instead of reaching `new(1 << 30)`. On the non-empty path I also reject headers whose `active_items` is inconsistent with `lg_cur_map_size`, or larger than the remaining bytes can hold, so the count can't drive an oversized `Vec::with_capacity` either. Added regression tests for all three; the existing C++/Java/Go round-trip fixtures still pass.

tisonkun (approved):

Issue #212, opened 2026-08-24
Implement Display for sketches


tisonkun:
Good point. I wonder if we'd have a tracking issue to `impl fmt::Display` for all sketches when applicible. This is what datasketches-java does IMO, but the format can be considered once more.

cc @notfilippo @ariesdevil @ZENOTME

_Originally posted by @tisonkun in https://github.com/apache/datasketches-rust/pull/204#discussion_r3814835517_

zhangxinyao88:
I'd like to work on this. To keep the first PR cohesive, I propose starting with the cardinality sketches: `HllSketch`, `HllUnion`, `CpcSketch`, and `CpcUnion`.

I'll implement summary-only `Display` output, using the Java summaries as a reference, and add tests for both empty and populated sketches. If that scope sounds good, I'll take it.

tisonkun:
Go ahead. I'm glad to see a demo to help decide.

tisonkun:
Thank @zhangxinyao88 for https://github.com/apache/datasketches-rust/pull/232.

This hints to me about one more issue:

In Rust, there is a loose contract:

1.  If a type implements both `FromStr` and `Display`, the following code snippet should pass:

```rust
assert_eq!(o, FromStr::from_str(o.to_string()))
```

2. The result of `Display` should be stable within one compatible version range.

However, the current printing format looks like:

```
HLL Sketch Summary:
  lg config k       : 12
  target type       : Hll8
  current mode      : List
  lower bound       : 0
  estimate          : 0
  upper bound       : 0
```

Seems more like a changable diagnostic report.

cc @leerho what is the design principle of these formats in datasketches-java. Are they mainly for debugging or diagnostics?

zhangxinyao88:
> @leerho what is the design principle of these formats in datasketches-java. Are they mainly for debugging or diagnostics?

[`HllSketch#toString()`](https://github.com/apache/datasketches-java/blob/main/src/main/java/org/apache/datasketches/hll/HllSketch.java#L87) is documented as a sketch summary, and [`CpcSketch#toString()`](https://github.com/apache/datasketches-java/blob/main/src/main/java/org/apache/datasketches/cpc/CpcSketch.java#L712) as a human-readable summary. They include state such as current mode and bounds.

I am inclined to expose a `summary()` method in Rust rather than implement `Display`, unless Java treats this text as a compatibility contract. Could you clarify whether that is the intended contract?

leerho:
[...]

PR #232, opened 2026-08-27
feat: add Display summaries for HLL and CPC

zhangxinyao88:
Part of #212.

## Summary

- implement `Display` for `HllSketch`, `HllUnion`, `CpcSketch`, and `CpcUnion`
- keep the output summary-only, without raw register or coupon details
- add coverage for empty and populated sketches and update the changelog

Example output:

```text
HLL Sketch Summary:
  lg config k       : 12
  target type       : Hll8
  current mode      : List
  lower bound       : 0
  estimate          : 0
  upper bound       : 0
```

```text
CPC Sketch Summary:
  flavor            : Empty
  lg k              : 11
  merged            : false
  lower bound       : 0
  estimate          : 0
  upper bound       : 0
  num coupons       : 0
```

## Testing

- `cargo x check`
- `cargo x test`
- `cargo +nightly clippy --tests --all-features --all-targets --workspace -- -D warnings`
- `cargo +nightly fmt --all --check`
- `cargo +nightly doc --package datasketches --all-features --no-deps`

### DIFF
--- a/datasketches/src/cpc/sketch.rs
+++ b/datasketches/src/cpc/sketch.rs
@@ -15,6 +15,7 @@
 // specific language governing permissions and limitations
 // under the License.
 
+use std::fmt;
 use std::hash::Hash;
 
 use crate::codec::SketchBytes;
@@ -469,6 +470,17 @@ impl CpcSketch {
                 .unwrap_or(0);
 
         size_of::<Self>() + heap_size
+    }
+}
+
+impl fmt::Display for CpcSketch {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        writeln!(f, "CPC Sketch Summary:")?;
+        writeln!(f, "  flavor            : {:?}", self.flavor())?;
+        writeln!(f, "  lg k              : {}", self.lg_k())?;
+        writeln!(f, "  merged            : {}", self.merge_flag)?;
+        writeln!(f, "  estimate          : {}", self.estimate())?;
+        writeln!(f, "  num coupons       : {}", self.num_coupons)
     }
 }
 
--- a/tests-integration/tests/cpc_test/display.rs
+++ b/tests-integration/tests/cpc_test/display.rs
@@ -1,0 +1,60 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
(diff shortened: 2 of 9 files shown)


datasketches/src/cpc/sketch.rs
```diff
@@ -471,6 +472,35 @@ impl CpcSketch {
...
 
+impl fmt::Display for CpcSketch {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        let flavor = match self.flavor() {
+            Flavor::Empty => "Empty",
+            Flavor::Sparse => "Sparse",
+            Flavor::Hybrid => "Hybrid",
+            Flavor::Pinned => "Pinned",
+            Flavor::Sliding => "Sliding",
+        };
```
tisonkun: I think you can either just use the `Debug` impl for flavor here or define a `pub const fn as_str` for Flavor`.

datasketches/src/cpc/sketch.rs
```diff
@@ -471,6 +472,35 @@ impl CpcSketch {
...
+
+        writeln!(f, "CPC Sketch Summary:")?;
+        writeln!(f, "  flavor            : {flavor}")?;
+        writeln!(f, "  lg k              : {}", self.lg_k())?;
+        writeln!(f, "  merged            : {}", self.merge_flag)?;
+        writeln!(
+            f,
+            "  lower bound       : {}",
+            self.lower_bound(NumStdDev::One)
```
tisonkun: Why `NumStdDev::One`? Could you point out the referenced implementation?

datasketches/src/cpc/sketch.rs
```diff
@@ -471,6 +472,35 @@ impl CpcSketch {
...
+        )?;
+        writeln!(f, "  estimate          : {}", self.estimate())?;
+        writeln!(
+            f,
+            "  upper bound       : {}",
+            self.upper_bound(NumStdDev::One)
```
tisonkun: Ditto

tests-integration/tests/cpc_test/display.rs
tisonkun, line 1: You may use `insta` for snapshot testing.

However, I have more to consider here now. Let me comment on the issue.

datasketches/src/cpc/sketch.rs
zhangxinyao88: Updated this to use the existing `Debug` implementation for `Flavor`.

datasketches/src/cpc/sketch.rs
zhangxinyao88: Removed this together with the lower bound, since the Java CPC summary does not include either bound.

datasketches/src/cpc/sketch.rs
zhangxinyao88: Good point. The Java HLL summary uses one standard deviation, but the Java CPC summary does not include bounds. I removed the CPC lower and upper bounds instead of introducing that choice here.

tests-integration/tests/cpc_test/display.rs
zhangxinyao88: Updated the exact `Display` assertions to use `insta` inline snapshots. I kept the populated cases as focused assertions so we do not snapshot estimator values unnecessarily.

zhangxinyao88:
@tisonkun, could you take another look when you have time? I addressed the review comments in `1c3e45b`.

tisonkun:
@zhangxinyao88 I left a comment at https://github.com/apache/datasketches-rust/issues/212#issuecomment-5447854830 that at least we may not implement this function as `Display`.

PR #233, opened 2026-08-27
perf(tdigest): fuse overlapping merge compression

tisonkun:
## Summary

- fuse sorted centroid merging with compression for overlapping, fully compressed inputs when T-Digest traverses them in descending order
- keep the existing two-pass linear merge for forward compression and disjoint ranges, and retain the stable-sort fallback for buffered or unsorted inputs
- extend the serialized partial-state benchmark to cover 8- and 64-row overlapping inputs and pin a 64-state native merge sequence to its existing serialized bytes

## Design Notes

T-Digest alternates compression direction. During descending compression, both sorted inputs can be consumed from the end while completed centroids are written backward into the unused tail of the left allocation. This avoids materializing the fully merged array and then scanning it again, without adding another allocation.

Forward compression would need to shift unread input or allocate a second buffer. Disjoint ranges also make the existing merge nearly a single-run copy. Those cases intentionally keep the simpler path, making this an adaptive optimization rather than a blanket replacement.

Stable ordering remains unchanged: the left side wins descending ties, which is the reverse of the existing ascending order where the right side precedes the left.

## Performance

Divan A/B runs compared `main` at `c3c08f9` with this branch on the same machine, using a two-second minimum and 300 samples. Times are medians.

| workload | main -> this PR | change |
| --- | ---: | ---: |
| merge: two overlapping 100k-row digests | 2.080 us -> 1.581 us | 24.0% faster |
| compute: 64 overlapping native states, 8 rows/state | 34.75 us -> 31.91 us | 8.2% faster |
| compute: 64 overlapping native states, 64 rows/state | 73.50 us -> 65.04 us | 11.5% faster |
| compute: 64 disjoint native states, 8 rows/state | 30.66 us -> 29.70 us | effectively neutral |
| compute: 64 disjoint native states, 64 rows/state | 64.50 us -> 63.91 us | effectively neutral |

Allocation counts and allocated bytes are unchanged. These local microbenchmarks are directional rather than a performance contract.

## Compatibility

- the public API and serialized format are unchanged
- the 64-state regression snapshot remains byte-for-byte identical to `main`
- buffered, unsorted, forward-order, and disjoint merges retain their previous paths

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,7 @@ All significant changes to this project will be documented in this file.
 
 ### Performance improvements
 
-* Reduce T-Digest allocation overhead and retained memory across updates, compression, merges, serialization, deserialization, and freezing; linearly merge sorted centroid buffers and decode validated native payloads directly while preserving the serialized format.
+* Reduce T-Digest allocation overhead and retained memory across updates, compression, merges, serialization, deserialization, and freezing; linearly merge sorted centroid buffers, fuse overlapping reverse-order merges with compression, and decode validated native payloads directly while preserving the serialized format.
 * Reduce CPC serialization and deserialization allocations by encoding directly into the output buffer and decoding directly from the input payload.
 
 ### Bug fixes
--- a/benchmarks/tdigest/merge.rs
+++ b/benchmarks/tdigest/merge.rs
@@ -118,11 +118,11 @@ fn serialized_partials(bencher: Bencher, rows_per_partial: usize) {
         });
 }
 
-#[divan::bench]
-fn serialized_overlapping_partials(bencher: Bencher) {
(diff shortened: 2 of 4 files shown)


tisonkun:
I opened #234 as a smaller alternative based on the same benchmark family. It keeps the existing eager path for larger or weighted inputs and limits the production change to bulk extension plus deferred compression for small unit-weight merges. I am leaving this draft open for now because its fused merge/compression path remains useful as a separate large overlapping-input experiment, but it is not required by #234.

tisonkun:
Closing this draft after reevaluating the tradeoff. The implementation already merged on main captures the larger general-purpose gain, while this overlap-specific path adds more maintenance cost than the measured incremental improvement currently justifies.

PR #234, opened 2026-08-27
perf(tdigest): streamline small partial aggregation

tisonkun:
## Summary

- implement `Extend<f64>` for bulk mutable T-Digest updates
- defer compression when merging at most 32 unit-weight centroids, so repeated small partial merges share one compression cycle
- keep weighted and larger inputs on the existing eager merge path to preserve its memory bound and overlapping-input behavior
- extend the public benchmarks around the batching boundary and bulk updates

## Design

A centroid with weight one is distribution-equivalent to one raw update for the next compression. A small all-unit-weight source can therefore be appended to the accumulator's unmerged tail without changing the serialized format or query semantics. The existing compression threshold still bounds the deferred work.

The 32-centroid cutoff is a performance and memory policy rather than a correctness boundary. Larger inputs retain eager compression because deferring them can accumulate a much larger temporary buffer, especially when sorted ranges overlap.

The specialized path is kept out of the eager merge loop under whole-program optimization. The buffer owns the ordering operation: source unmerged values precede its compressed prefix, matching the stable tie order of the existing eager merge.

## Benchmarks

64 serialized partials, median of 5-second Divan runs on Apple Silicon:

| rows / partial | disjoint before | disjoint after | overlapping before | overlapping after |
| ---: | ---: | ---: | ---: | ---: |
| 8 | 28.23 us | 4.90 us | 33.32 us | 7.86 us |
| 32 | 46.74 us | 15.97 us | 53.32 us | 32.02 us |
| 64 | 62.61 us | 62.20 us | 70.90 us | 71.24 us |

The eager 64-centroid path stays within about 0.5% in both directions. Peak live allocation grows from 4.2 KB to 16.4 KB for 8-centroid inputs and from 4.6 KB to 59.9 KB for 32-centroid inputs; the 64-centroid path remains at 9.2 KB.

For a contiguous update stream, `Extend<f64>` measured 12.52 us versus 12.56 us for 1,000 scalar updates and 2.239 ms versus 2.277 ms for 100,000 scalar updates. Its larger purpose is to give iterator-based callers one bulk boundary without forcing `update` to be inlined at every call site.

## Validation

- `cargo x check`
- `cargo x test`
- `cargo x lint`
- `cargo x bench`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,7 @@ All significant changes to this project will be documented in this file.
 
 ### Performance improvements
 
-* Reduce T-Digest allocation overhead and retained memory across updates, compression, merges, serialization, deserialization, and freezing; linearly merge sorted centroid buffers and decode validated native payloads directly while preserving the serialized format.
+* Reduce T-Digest allocation overhead and retained memory across updates, compression, merges, serialization, deserialization, and freezing; support bulk extension, defer compression for small unit-weight merges, linearly merge larger sorted centroid buffers, and decode validated native payloads directly while preserving the serialized format.
 * Reduce CPC serialization and deserialization allocations by encoding directly into the output buffer and decoding directly from the input payload.
 
 ### Bug fixes
--- a/benchmarks/tdigest/merge.rs
+++ b/benchmarks/tdigest/merge.rs
@@ -102,7 +102,7 @@ fn partials(bencher: Bencher) {
         });
 }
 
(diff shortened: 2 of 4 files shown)


tisonkun:
Closing this draft after reevaluating the tradeoff. The small-state batching result is workload-sensitive and increases temporary memory. We should first establish that tiny unit-weight merges dominate a representative workload before adding this policy.

PR #235, opened 2026-08-28
perf(tdigest): stream sorted merges through compression

tisonkun:
## Summary

- stream overlapping sorted centroid buffers directly through T-Digest compression during reverse-order merges
- centralize the compression decision in one state machine shared by the regular and fused paths
- retain the existing linear two-pass path for forward-order and disjoint merges, plus the stable-sort fallback for buffered or unsorted inputs
- preserve serialized output byte-for-byte for a representative 64-state merge sequence

This is an alternative implementation of the optimization explored in #233. Its main design difference is that the compression rule has a single owner instead of being duplicated inside the fused merge loop.

## Design Notes

T-Digest alternates compression direction. During descending compression, two sorted inputs can be consumed from their ends while completed centroids are written backward into the unused tail of the left allocation. The output cannot overwrite unread left-hand input because the number of completed centroids never exceeds the number of consumed inputs.

`CentroidCompression` owns the scale-function boundary and the current centroid. The ordinary in-place compression loop and the fused two-way merge only provide the next centroid and store each completed result. Stable ordering is unchanged: descending traversal takes the left side on ties, which reverses back to the existing right-before-left ascending order.

Forward compression would overwrite unread input without another buffer. Disjoint sorted runs are also cheaper on the existing linear materialize-then-compress path. Those cases deliberately remain unchanged.

## Performance

Divan A/B runs compare `main` at `c3c08f9` with this branch on the same machine. Times are medians of five-second runs over 64 serialized native states.

| workload | main -> this PR | change |
| --- | ---: | ---: |
| 8 rows/state, overlapping ranges | 34.18 us -> 30.56 us | 10.6% faster |
| 64 rows/state, overlapping ranges | 72.35 us -> 63.36 us | 12.4% faster |
| 8 rows/state, disjoint ranges | 29.35 us -> 28.86 us | effectively neutral |
| 64 rows/state, disjoint ranges | 62.98 us -> 63.15 us | effectively neutral |

Allocation counts and serialized formats are unchanged. These local microbenchmarks are directional rather than a performance contract.

## Compatibility

- no public API or serialized-format changes
- native serialized merge regression remains byte-for-byte identical to `main`
- buffered, unsorted, forward-order, and disjoint merges keep their prior paths

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,7 @@ All significant changes to this project will be documented in this file.
 
 ### Performance improvements
 
-* Reduce T-Digest allocation overhead and retained memory across updates, compression, merges, serialization, deserialization, and freezing; linearly merge sorted centroid buffers and decode validated native payloads directly while preserving the serialized format.
+* Reduce T-Digest allocation overhead and retained memory across updates, compression, merges, serialization, deserialization, and freezing; linearly merge sorted centroid buffers, fuse overlapping reverse-order merges with compression, and decode validated native payloads directly while preserving the serialized format.
 * Reduce CPC serialization and deserialization allocations by encoding directly into the output buffer and decoding directly from the input payload.
 
 ### Bug fixes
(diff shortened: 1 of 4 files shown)


tisonkun:
Closing this alternative after reevaluating the tradeoff. The implementation already merged on main captures the larger general-purpose gain, while this overlap-specific path adds more maintenance cost than the measured incremental improvement currently justifies.

PR #238, opened 2026-08-28
refactor: unify fallible construction APIs

tisonkun:
## Summary

Use one naming rule for construction APIs: if caller-supplied configuration can fail, the ordinary method returns `Result`; infallible entry points continue to return the value directly. This removes the parallel `new`/`try_new` vocabulary and lets callers propagate construction failures without guessing which spelling is fallible.

Builders remain infallible while configuration is assembled and validate only when `build()` is called. Types whose standard configuration is known to be valid continue to provide an infallible `Default` implementation.

The pull request also rewrites the current `Unreleased` changelog around observable user impact and records the same user-facing changelog guidance in `CONTRIBUTING.md` and `AGENTS.md`.

## Resulting API shape

| Family / public type | Construction entry point(s) | Failure boundary | Notes |
| --- | --- | --- | --- |
| `BloomFilter` | `BloomFilterBuilder::with_accuracy(...)` or `with_size(...)` | `build() -> Result<_, Error>` | Builder acquisition and setters remain infallible; validation stays deferred to `build`. |
| `CountMinSketch` | `new(...)` or `with_seed(...)` | Both return `Result` | `suggest_num_buckets` and `suggest_num_hashes` also return `Result` and enforce their individual constructor limits; the constructor still validates the combined table size. |
| `CpcSketch` / `CpcUnion` | `Default::default()`, `new(...)`, or `with_seed(...)` | `Default` is infallible; `new` and `with_seed` return `Result` | `CpcWrapper::new(bytes)` was already fallible and remains unchanged. |
| `FrequentItemsSketch` | `new(max_map_size)` | Returns `Result` | Invalid or oversized map sizes no longer panic. |
| `HllSketch` / `HllUnion` | `new(...)` | Returns `Result` | Invalid `lg_k` values no longer panic. |
| `ReqSketch` / `ReqUnion` | `Default::default()` or `new(...)` | `Default` is infallible; `new` returns `Result` | Removes `try_new`. |
| `TDigestMut` | `Default::default()` or `new(k)` | `Default` is infallible; `new` returns `Result` | Removes `try_new`; immutable `TDigest` is still obtained via `freeze`. |
| `ThetaSketch` / `ThetaUnion` | `ThetaSketchBuilder::default()` / `ThetaUnionBuilder::default()` | `build() -> Result<_, Error>` | There is no zero-argument builder `new()` and no `Sketch::builder()` shortcut; setters, including `seed`, remain infallible until `build`. |
| Theta intersection / A-not-B / Jaccard | `Default::default()` or `with_seed(...)` | `Default` is infallible; `with_seed` returns `Result` | The default-seed path remains convenient while custom seeds are validated explicitly. |
| `TupleSketch` / `TupleUnion` | `Builder::new(policy)` | Builder acquisition is infallible; `build()` returns `Result` | The required policy is supplied up front; all configuration validation remains at `build`. |
| `TupleIntersection` | `new(policy)` or `with_seed(policy, seed)` | `new` is infallible; `with_seed` returns `Result` | `new` uses the known-valid default seed. |
| Tuple A-not-B / Jaccard | `Default::default()` or `with_seed(...)` | `Default` is infallible; `with_seed` returns `Result` | These operators do not require a summary policy. |

## Seed errors

The internal seed-hash helper now takes the error kind from its caller:

- direct construction and builder validation report a reserved zero seed hash as `InvalidArgument`;
- deserialization reports the same condition as `InvalidData`, matching other serialized seed-compatibility failures;
- known-valid defaults and reconstruction from already validated state remain infallible internal invariants.

Theta hash tables retain the seed hash validated during construction, and deserializers compute an expected seed hash once before dispatching to format-specific decoding. This keeps later access infallible and avoids repeating the hash calculation; it is primarily an error-boundary change rather than a performance feature.

Direct custom-seed constructors consistently use `with_seed(...)`. Builders keep an infallible `seed(...)` setter and surface an invalid seed from `build()`, avoiding a growing family of `try_with_*` names as more configuration parameters are added.

## Count-Min parameter suggestions

- `relative_error` must be finite and greater than zero;
- the suggested bucket count is at least the constructor minimum of three and errors when the requested target exceeds the supported table limit;
- zero confidence produces the minimum valid hash count of one;
- cross-field `num_hashes * num_buckets` validation remains in the constructor.

## Validation

- `cargo x check`
- `cargo x test`
- `cargo x lint`

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -5,3 +5,5 @@ Before planning or modifying this repository, read [CONTRIBUTING.md](CONTRIBUTIN
 For test changes, pay particular attention to the "Integration test layout" and "Serialization snapshots" sections. Keep the documented workflow synchronized with structural changes, and run the applicable `cargo x check`, `cargo x test`, and `cargo x lint` commands before handing work back.
 
 Apply the changelog guidance in [CONTRIBUTING.md](CONTRIBUTING.md) to every change. Update the permanent `Unreleased` section in the same pull request for significant user-visible behavior, and do not add entries mechanically for excluded maintenance work.
+
+Treat `CHANGELOG.md` as release notes for users rather than a summary of implementation work. Name the affected API or workload and the observable outcome, and keep performance claims within the scenario supported by evidence.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@ All significant changes to this project will be documented in this file.
 
 ### Bug fixes
 
+* Count-Min parameter suggestions now return constructor-valid values and reject relative-error targets that require more buckets than the sketch supports.
 * Bloom filter deserialization now rejects malformed images with inconsistent counts or payload lengths, while valid images with a dirty cached count are restored correctly.
 * `FrequentItemsSketch` now enforces the cross-language map-size limit of `2^30` consistently. Oversized construction returns `InvalidArgument`, and malformed or oversized serialized images return `InvalidData` instead of panicking or attempting excessive allocation.
 * T-Digest compression now supports `k = u16::MAX` without overflowing.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,8 +6,9 @@ All significant changes to this project will be documented in this file.
 
 ### Breaking changes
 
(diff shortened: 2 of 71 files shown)


tisonkun:
fyi @ariesdevil @notfilippo @pmcgleenon

PR #242, opened 2026-08-28
refactor(req): remove ReqUnion

### LINKED ISSUE #217: Review ReqUnion
cc @AlexanderSaydakov @leerho You can review the `ReqUnion` design here and give some high-level design comments.

_Originally posted by @tisonkun in https://github.com/apache/datasketches-rust/pull/204#discussion_r3814852391_

cc @pmcgleenon may you share more background of this struct since it seems not to be included in the datasketches-java or datasketches-cpp library already. And if it's valueable, we may bring it back to other impls.

tisonkun:
## Summary

- remove `ReqUnion`, which only delegated to an inner `ReqSketch`
- keep `ReqSketch::merge` as the REQ aggregation API and add a usage example
- update the README and unreleased changelog to describe merging without a separate union type

## Rationale

The Java, C++, Python, and Go REQ implementations expose aggregation directly through the sketch's `merge` method. Unlike CPC, HLL, and Theta unions, `ReqUnion` maintained no distinct accumulator state or configuration-reconciliation behavior, and `to_sketch` cloned the wrapped sketch.

REQ has not been included in a release yet, so the discarded intermediate API does not require a migration entry. This change intentionally leaves the existing `ReqSketch::merge` compatibility rules, including mixed-`k` behavior, unchanged.

## Testing

- `cargo x check`
- `cargo x test`
- `cargo x lint`

Closes #217.


tisonkun:
@pmcgleenon Seems ReqUnion is almost identical to continuous calls to `ReqSketch::merge`. I tend to reduce concepts from users and add them only if needed or if concrete use cases occur.

tisonkun:
... to be clear, a merge is not a final decision. You can always do review after commits and revert a commit with a reason.

PR #250, opened 2026-08-30
fix(tdigest): interpolate quantiles toward the nearer centroid and normalize the left tail

jaideeppyne:
## Summary

- Correct the interior quantile interpolation weights so the estimate moves toward the nearer centroid as rank increases.
- Fix the mirrored tail formulas: subtract from `max` in the right quantile tail and normalize the left rank tail by total weight.
- Handle the weight-two right-tail boundary without evaluating an indeterminate `0 / 0` expression.
- Add focused exact regressions and monotonicity properties for directly updated and merged digests.
- Document the bug fix in the changelog.

## Why

The interior interpolation paired each centroid mean with the distance from that same side of the bracket. Linear interpolation needs the opposite-side distance, so the previous result could move toward the lower centroid as rank increased and make quantiles non-monotonic.

The left and right tail branches contained two independent mirror errors. The right-tail formula also degenerates when the last centroid has weight two: at the only boundary that reaches the formula, both its numerator and denominator are zero. Returning the known maximum at that boundary keeps the handling local and avoids changing interpolation elsewhere.

## Validation

- `cargo x check`
- `cargo x test`
- `cargo x lint`
- Exact interior, tail, and weight-two boundary regressions
- QuickCheck monotonicity properties for digests built through both `update` and `merge`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -47,6 +47,8 @@ All significant changes to this project will be documented in this file.
 * HLL, Theta, and Tuple deserializers now return `InvalidData` for malformed payload sizes and entry counts instead of risking oversized allocations or decoding failures.
 * Malformed CPC images now return `InvalidData` instead of panicking.
 * Seeded deserializers now return `InvalidData` rather than panicking when the caller supplies a seed whose hash is the reserved zero value.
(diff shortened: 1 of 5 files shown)


tisonkun (approved):
Thanks for your contribution! Merging ...

PR #254, opened 2026-08-31
fix(theta,tuple): canonicalize empty sampled sketch state

jaideeppyne:
## Problem

A Theta-family update sketch configured with `sampling_probability < 1.0` needs to represent two different facts:

- its operational retention theta, used to screen future updates, is below `MAX_THETA`;
- while the sketch is empty, its public theta must be `MAX_THETA` and it must not report estimation mode.

The previous representation stored emptiness, theta, retained entries, and ordering as independent fields. That allowed contradictory combinations such as `empty = true` with `theta < MAX_THETA`.

For an empty sampled sketch this produced:

```text
theta64 < MAX_THETA
theta == sampling_probability
is_estimation_mode == true
```

It also affected set-operation results. In particular, A-not-B could copy an empty sampled input into an empty result that still carried the sampling theta. The result serialized with an explicit theta, but deserialization discarded it because the empty flag was set, so serialization was not stable across a round trip.

Java and C++ keep the sampling threshold internally but mask public theta while the sketch is empty. They report `theta64 = MAX_THETA`, `theta = 1.0`, and `is_estimation_mode = false` in this state.

## Design

This change separates operational hash-table state from the state represented by a sketch:

- `SketchHashTable` owns only `retention_theta`, the threshold used to retain future updates. It no longer owns sketch emptiness.
- Mutable Theta and Tuple sketches own their `is_empty` state. An update call makes the sketch non-empty even when theta screens out the hash, so a non-empty sketch may legitimately retain zero entries.
- `CompactSketchState<E>` represents compact sketches as either `Empty { seed_hash }` or `NonEmpty { retained_entries, theta, seed_hash, ordered }`. The empty variant cannot carry a stale theta, retained entries, or a contradictory ordering flag.
- `ThetaFamilySketchMetadata` applies the same distinction at the set-operation boundary. Empty inputs expose only their seed hash; theta, ordering, and retained count exist only for non-empty inputs.
- Intersection keeps its necessary `Uninitialized` / `Empty` / `NonEmpty` state machine. Union uses a local optional result theta, where `None` means that no non-empty input has been incorporated.

The state types live in `thetafamily/common/sketch_state.rs`, and callers import them from that owning module rather than through an internal facade re-export.

## Observable behavior

- Empty Theta and Tuple update sketches report `theta64 = MAX_THETA`, `theta = 1.0`, and `is_estimation_mode = false` for every valid sampling probability.
- After an update is attempted, the configured sampling theta becomes visible even if zero entries were retained.
- `compact()` preserves the distinction between a truly empty sketch and a non-empty sketch with zero retained entries.
- Empty union, intersection, and A-not-B results use the canonical ordered compact representation and survive serialization round trips without changing state.
- Non-empty sketch and serialization behavior is unchanged.

## Tests

The regression coverage now focuses on observable state transitions rather than repeating the same assertions over probability matrices:

- empty sampled sketch -> screened update -> compact -> reset, for both Theta and Tuple;
- canonical empty union output before updates and after reset;
- canonical empty A-not-B and intersection results;
- serialization round trips for canonical empty compact sketches;
- preservation of logically non-empty sketches with zero retained entries.

Validated locally with:

- `cargo x check`
- `cargo x test`
- nightly Clippy, rustfmt, and rustdoc checks with warnings denied

AI assistance: Claude Code assisted the initial investigation and implementation. Codex assisted the subsequent review, state-model refactor, naming and test consolidation. The resulting changes were reviewed and validated by the authors.


datasketches/src/thetafamily/common/sketch_state.rs
```diff
@@ -0,0 +1,270 @@
...
+    pub fn new(value: u64) -> Self {
+        Self::try_new(value)
+            .unwrap_or_else(|| panic!("theta must be in [1, {MAX_THETA}], got {value}"))
+    }
+
+    pub fn get(self) -> u64 {
+        self.0.get()
+    }
+
+    pub fn is_estimation_mode(self) -> bool {
```
ZENOTME: I'm not sure whether `is_estimation_mode` in ThetaThreshold is a correct abstract. 🤔 It can be a method in ThetaSketchState. But ThetaThreshold is more like a "value".

tisonkun: `ThetaThreshold` is to be removed.

Local Codex is woring now ...

tisonkun:
@ZENOTME You may do a post-merge review and fix any regression or apply more ideas.

PR #264, opened 2026-09-02
refactor: remove the rand runtime dependency and add common random bit function

### LINKED ISSUE #203: missing common_random support in library

jeffoodchain:
## Summary

Remove the `rand` runtime dependency from the published `datasketches` crate and add our own random generator.

KLL and REQ only require a random boolean when choosing which half of a
compaction range to promote. This PR replaces those calls with a small,
crate-private `random_bit()` implementation backed by a thread-local
SplitMix64 generator.

Closes #203 

## Changes

- add a crate-private, thread-local SplitMix64 random-bit generator
- seed each thread's generator on first use through `RandomState`
- use `random_bit()` for KLL and REQ compaction decisions
- remove `rand` from the `kll` and `req` feature dependency lists
- remove `rand` from the published library's normal dependencies
- document the dependency change in the changelog

The workspace still uses `rand` for benchmarks, but downstream users of the
`datasketches` crate no longer pull it in.

## Compatibility

This does not change the public API or serialization format.

KLL and REQ compaction remain randomized. The new generator only replaces the
source of the random coin used to select which items are promoted.

## Testing

- `cargo x check`
- `cargo x lint`
- `cargo x test`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,14 +15,11 @@ All significant changes to this project will be documented in this file.
 ### Improvements
 
 * Improve truncated-input diagnostics across sketch deserializers.
+* The crate no longer has any runtime dependencies. The `kll` and `req` features previously pulled in `rand`; compaction now draws its coin from an in-tree generator.
 
 ### Bug fixes
 
 * T-Digest deserialization now rejects unknown or conflicting flags, reversed extrema, out-of-range values, unsorted centroids, and non-empty images without stored values.
-
-### Notable changes
-
-* The crate no longer has any runtime dependencies. The `kll` and `req` features previously pulled in `rand`; compaction now draws its coin from an in-tree generator.
 
 ## v0.5.0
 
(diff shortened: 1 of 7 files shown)


tisonkun (approved):
Thanks!

PR #271, opened 2026-09-03
feat(bloom): introduce BloomFilterInvertedView for inverted queries

### LINKED ISSUE #270: BloomFilter: make post-invert semantics observable in the type system
## Context

`BloomFilter::invert()` flips every bit in the backing array. #194 raised the resulting contract questions and #202 documented the altered guarantees, but left the API-shape decision open: after inversion the type signature is unchanged and still permits every update operation.

The semantics after inversion are worth pinning down precisely. Let `B` be the bit array before inversion and `S(x)` the `k` hash positions of an item `x`:

- `contains(x) == true` after inversion means every position of `S(x)` was `0` in `B`, so `x` was **definitely not** inserted before the inversion. This direction is certain; it has no error probability.
- `contains(x) == false` means at least one position of `S(x)` was `1` in `B`. All inserted items land here, but so does any absent item with at least one colliding position — with probability roughly `1 - (1 - load_factor)^k`, which is close to 1 for typical configurations. So `false` carries little information.
- Inversion is therefore **not** the exact logical complement of the pre-inversion answer: a `true` result requires *all* `k` positions to have been clear, not merely one.

For updates after inversion:

- `insert(z)` only sets bits, so it can only flip query results from `false` to `true` — it manufactures new "definitely absent" claims. The claim about `z` itself is sound only if `z` was absent before inversion, a precondition the filter cannot check. Existing `true` answers are never corrupted, since setting bits preserves all-set patterns.
- The opposite update — recording that a newly seen item belongs to the original stream — would require *clearing* bits, i.e. the deletion a Bloom filter fundamentally cannot support.
- `union()`/`intersect()` remain well-defined as raw bit-array boolean operations and, combined with `invert()`, enable compositions such as approximate set difference (`A AND NOT B`). `invert()` is also an exact involution: inverting twice restores the original filter bit-for-bit, guarantees included.

Nothing in the object records that it was inverted, and the serialized format stores only the bit array, so invertedness cannot be persisted without a format change. Any marker is necessarily ephemeral and type-level. This matches the C++ and Java implementations, where `invert()` is a plain mutating bit operation:

- C++: https://github.com/apache/datasketches-cpp/blob/master/filters/include/bloom_filter.hpp
- Java: https://github.com/apache/datasketches-java/blob/main/src/main/java/org/apache/datasketches/filters/bloomfilter/BloomFilter.java

## Problem

After `invert()`, a `BloomFilter` is a valid bit array but no longer satisfies the update-side contract of a set-membership sketch: `insert()` and `contains_and_insert()` assert a "definitely absent" fact whose soundness precondition the filter cannot verify. Queries remain well-defined (with the flipped, one-sided meaning above), but nothing in the type system distinguishes the two modes, so unsound updates compile silently.

### A. Documentation only (status quo)

Keep `invert(&mut self)` and extend the post-invert qualification to the update methods: state that `insert()` after inversion asserts a caller-guaranteed absence, and that `union()`/`intersect()` compose raw bit state. Non-breaking and consistent with C++/Java, but the footgun remains — nothing prevents or flags unsound updates.

### B. Consuming invert returning a read-only view (recommended)

```rust
impl BloomFilter {
    pub fn invert(self) -> InvertedBloomFilter;
}

impl InvertedBloomFilter {
    /// `true` means the item was definitely not inserted before inversion.
    pub fn contains<T: Hash>(&self, item: &T) -> bool;
    // capacity(), num_hashes(), seed(), bits_used(), load_factor()

    /// Exact round-trip: restores the original filter bit-for-bit.
    pub fn invert(self) -> BloomFilter;

    /// Explicit escape hatch for boolean composition (e.g. A AND NOT B);
    /// the caller takes responsibility for the bit-level interpretation.
    pub fn into_filter(self) -> BloomFilter;
}
```

- Makes the altered contract observable at compile time: query-only after inversion, with `insert()`/`contains_and_insert()` unrepresentable.
- Preserves the legitimate uses: the exact double-invert round-trip and boolean composition both remain possible, the latter through an explicit, documented conversion.
- Wire format unchanged: an inverted filter serializes as its raw bits and deserializes as a plain `BloomFilter`. If `InvertedBloomFilter` exposes `serialize()`, that asymmetry must be documented.

Cost: a breaking change to `invert()`'s signature, and an API-level (not format-level) divergence from C++/Java.

### C. Runtime flag

Track an `inverted` boolean and panic or no-op on `insert()` after inversion. Adds per-object state, cannot round-trip through serialization without a format change, and turns a compile-time question into a runtime failure. Not recommended.

## Desired outcome

- Decide between A and B (B recommended); C is recorded as rejected.
- If B: add `InvertedBloomFilter` under `datasketches/src/bloom/`, change `invert()` to consume `self`, move the inverted-semantics documentation onto the new type, and update the doctests in `datasketches/src/bloom/mod.rs` and `sketch.rs`.
- Update `tests-integration/tests/bloom_test/sketch.rs`: `test_invert_is_reversible` becomes an `invert().invert()` round-trip; add tests covering the flipped one-sided query meaning (inserted items always return `false`; `true` implies definite absence) and `into_filter()` composition.
- Add a changelog entry per `CONTRIBUTING.md`.
- Run `cargo x check`, `cargo x test`, and `cargo x lint` before submitting.

v0id-X:
- Replaces in-place bit inversion on BloomFilter with an owned BloomFilterInvertedView.
- 
- Prevents unsound mutations on an inverted filter while retaining membership queries.
- 
- Provides bi-directional conversion via invert() and into_filter().
- 
- Added unit and integration tests covering the new view type and conversion paths.

Closes #270

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,14 +3,16 @@
 All significant changes to this project will be documented in this file.
 
 ## Unreleased
-- feat(bloom): make post-invert semantics observable via `BloomFilterInvertedView` (#270, #271)
+
 
 ### Breaking changes
 
+* `BloomFilter::invert` now consumes `self` and returns a read-only `BloomFilterInvertedView` instead of mutating in place. Call `invert()` on the view to restore the original updatable filter.
 * Move `SearchCriteria` from `req` to `common` and remove its `Default` implementation. Import `datasketches::common::SearchCriteria` and explicitly choose `Inclusive` or `Exclusive` for each query.
 
 ### New features
 
+* Add `BloomFilterInvertedView` to represent inverted Bloom filters with read-only query semantics.
 * Add KLL sketches behind the `kll` feature, with rank, quantile, PMF, and CDF queries, merging, totally ordered custom item types, a `KllFloat` adapter for non-NaN floating-point values, and serialization.
 
 ### Improvements
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 All significant changes to this project will be documented in this file.
 
 ## Unreleased
+- feat(bloom): make post-invert semantics observable via `BloomFilterInvertedView` (#270, #271)
 
 ### Breaking changes
 
--- a/datasketches/src/bloom/sketch.rs
+++ b/datasketches/src/bloom/sketch.rs
@@ -39,7 +39,7 @@ const EMPTY_FLAG_MASK: u8 = 1 << 2;
 /// * Tunable false positive rate
 /// * Constant space usage
 ///
-/// These guarantees hold until [`invert()`](Self::invert) is called; see its documentation.
+/// These guarantees hold unless inverted via [`invert()`](Self::invert).
 #[derive(Debug, Clone, PartialEq)]
 pub struct BloomFilter {
     /// Hash seed for all hash functions
@@ -251,13 +251,17 @@ impl BloomFilter {
         Ok(())
     }
 
-    /// Inverts all bits in the filter.
+    /// Consumes the filter and inverts all its bits, returning a read-only inverted view.
     ///
     /// This approximately inverts the notion of set membership. After inversion, neither the
     /// no-false-negative nor the false-positive guarantee holds: inserted items may return
-    /// `false` from [`contains()`](Self::contains), and [`is_empty()`](Self::is_empty),
-    /// [`bits_used()`](Self::bits_used), and [`load_factor()`](Self::load_factor) describe the
-    /// raw bit state rather than the inserted items.
+    /// `false` from [`contains()`](BloomFilterInvertedView::contains), and metadata methods
+    /// describe the raw inverted bit state.
+    ///
+    /// Updates are disallowed on an inverted view to prevent unsound filter states. An inverted
+    /// view can be converted back into an updatable [`BloomFilter`] via
+    /// [`invert()`](BloomFilterInvertedView::invert) or
+    /// [`into_filter()`](BloomFilterInvertedView::into_filter).
     ///
     /// # Examples
     ///
@@ -269,20 +273,25 @@ impl BloomFilter {
     ///     .unwrap();
     /// filter.insert("apple");
     ///
-    /// filter.invert();
-    /// // Now "apple" probably returns false, and most other items return true
+    /// let inverted = filter.invert();
+    /// // "apple" likely returns false in the inverted view:
+    /// assert!(!inverted.contains(&"apple"));
+    ///
+    /// // Inverting back restores the original filter state:
+    /// let restored = inverted.invert();
+    /// assert!(restored.contains(&"apple"));
     /// ```
-    pub fn invert(&mut self) {
+    pub fn invert(mut self) -> BloomFilterInvertedView {
         for word in &mut self.bit_array {
             *word = !*word;
         }
         self.num_bits_set = self.capacity() as u64 - self.num_bits_set;
+        BloomFilterInvertedView { inner: self }
     }
 
     /// Returns whether no bits are set in the filter.
     ///
-    /// In normal operation this means no items were inserted. After [`invert()`](Self::invert),
-    /// it reports the raw bit state instead.
+    /// Returns `true` if no bits are set in the filter.
     pub fn is_empty(&self) -> bool {
         self.num_bits_set == 0
     }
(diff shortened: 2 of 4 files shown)


datasketches/src/bloom/sketch.rs
```diff
@@ -39,7 +39,7 @@ const EMPTY_FLAG_MASK: u8 = 1 << 2;
 /// * Tunable false positive rate
 /// * Constant space usage
 ///
-/// These guarantees hold until [`invert()`](Self::invert) is called; see its documentation.
+/// These guarantees hold unless inverted via [`invert()`](Self::invert).
```
tisonkun: This is no longer relevant because the inverted view is now a brand-new read-only view. It is no longer a BloomFilter.

datasketches/src/bloom/sketch.rs
```diff
@@ -251,13 +251,17 @@ impl BloomFilter {
...
-    /// `false` from [`contains()`](Self::contains), and [`is_empty()`](Self::is_empty),
-    /// [`bits_used()`](Self::bits_used), and [`load_factor()`](Self::load_factor) describe the
-    /// raw bit state rather than the inserted items.
+    /// `false` from [`contains()`](BloomFilterInvertedView::contains), and metadata methods
+    /// describe the raw inverted bit state.
+    ///
+    /// Updates are disallowed on an inverted view to prevent unsound filter states. An inverted
+    /// view can be converted back into an updatable [`BloomFilter`] via
+    /// [`invert()`](BloomFilterInvertedView::invert) or
+    /// [`into_filter()`](BloomFilterInvertedView::into_filter).
```
tisonkun: I think this can be a document of `BloomFilterInvertedView` rather than details on the method.

datasketches/src/bloom/sketch.rs
```diff
@@ -613,6 +622,114 @@ impl BloomFilter {
...
+    /// inverting the bits again.
+    ///
+    /// Equivalent to [`invert()`](Self::invert).
+    pub fn into_filter(mut self) -> BloomFilter {
+        for word in &mut self.inner.bit_array {
+            *word = !*word;
+        }
+        self.inner.num_bits_set = self.inner.capacity() as u64 - self.inner.num_bits_set;
+        self.inner
+    }
```
tisonkun: I don't think we need alias here. Please think of it by yourself rather than just blindly follow the issue description.

datasketches/src/bloom/sketch.rs
```diff
@@ -613,6 +622,114 @@ impl BloomFilter {
...
+
+    /// Returns a reference to the underlying [`BloomFilter`] representation.
+    pub fn as_filter(&self) -> &BloomFilter {
+        &self.inner
+    }
```
tisonkun: Why do we have this method?

CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 All significant changes to this project will be documented in this file.
 
 ## Unreleased
+- feat(bloom): make post-invert semantics observable via `BloomFilterInvertedView` (#270, #271)
```
tisonkun: Please follow the changelog format like other entries.

datasketches/src/bloom/sketch.rs
v0id-X: Yes, Removing it.

datasketches/src/bloom/sketch.rs
v0id-X: Agreed. I'll trim the method documentation down to its core behavior and move the details regarding disallowed mutations and reinversion up to the BloomFilterInvertedView struct documentation.

datasketches/src/bloom/sketch.rs
v0id-X: That makes a lot more sense. Symmetrical .invert() (filter.invert().invert()) is way cleaner anyway. Dropping into_filter and keeping just invert(self) -> BloomFilter.

datasketches/src/bloom/sketch.rs
v0id-X: I thought about it, this makes the abstraction weaker risking misuse. Removing it entirely.

CHANGELOG.md
v0id-X: My bad, will fix the formatting to match the rest of the file.

v0id-X:
All review comments addressed and updated:
- Removed the outdated guarantee disclaimer on `BloomFilter`.
- Moved mutation/inversion details from `BloomFilter::invert`'s doc to `BloomFilterInvertedView`'s struct-level documentation.
- Removed the `into_filter` alias in favor of symmetrical `invert(self) -> BloomFilter`.
- Removed `as_filter` to preserve view encapsulation.
- Aligned `CHANGELOG.md` under `### Breaking changes` and `### New features` per `CONTRIBUTING.md`.
- Uncommented the inverted view doctest example in `bloom/mod.rs`.

tisonkun:
Thank you for picking this up — the implementation here was clean and well-tested. After a closer look at the underlying semantics, though, we've decided not to expose the inverted state at all, which makes this approach moot.

The reasoning, in short: an inverted Bloom filter only answers "definitely not inserted" on its `true` side, and the original filter already answers that (with a much higher hit rate) on its `false` side — so a read-only inverted view is a strictly worse absence oracle than the filter it came from. The one composition with a clean per-item meaning is A AND NOT B, and that use case is better served by an explicit set-difference operation than by exposing an inverted intermediate. #278 implements that: it removes `invert()` and adds `BloomFilter::difference` with documented semantics.

Closing this in favor of #278; the cross-language discussion is tracked at https://github.com/apache/datasketches-java/issues/766. Your exploration here was still useful input for reaching that conclusion.

PR #272, opened 2026-09-03
build: drop the rust-toolchain.toml pin

tisonkun:
## Problem

`cargo x lint` auto-installs taplo-cli via an unpinned `cargo install taplo-cli` when taplo is not on PATH (`xtask/src/main.rs`, `ensure_installed`). Recent taplo-cli releases pull in transitive dependencies (`icu_*`, `time`) that require rustc 1.88, so the install now fails on a clean environment where the pinned 1.86.0 toolchain from `rust-toolchain.toml` is active.

## Change

Drop `rust-toolchain.toml` entirely:

- The workspace already declares its MSRV with `rust-version` in `Cargo.toml`, and CI verifies it through the test matrix, so the pin is redundant for correctness.
- Every CI job already deleted `rust-toolchain.toml` before installing its toolchain, meaning the pin only affected local development — where it broke `cargo install` of lint tools. The now-obsolete deletion steps are removed from CI.
- `CONTRIBUTING.md` setup instructions are updated: any toolchain at or above the MSRV works for build/test, and a nightly toolchain is required for `cargo x lint` / `cargo x check`.

Verified locally with `cargo x lint` and `cargo x check`.


tisonkun:
@freakyzoidberg This should resolve the issue you pointed out in the release vote thread.

If the experience still has some issue, I'd either try to improve the error message or pin a version for tool. So far, I tend to use latest dev tools to use their nightly features.

PR #273, opened 2026-09-03
fix: unify sketch query semantics

### LINKED ISSUE #266: Define consistent fallible semantics for sketch query APIs
## Context

The Rust implementations currently use different error models for distribution queries:

- T-Digest returns `Option` when querying an empty sketch and panics for invalid ranks, NaN rank inputs, or invalid split points.
- KLL and REQ return `Result`, but currently classify an empty sketch as `InvalidArgument` alongside malformed query arguments.
- Deserializers separately report malformed serialized state as `InvalidData`.

An empty sketch is a valid object state: it can be created, serialized, merged, and updated. However, quantile, rank, CDF, and PMF values are undefined until the sketch has observed data. That condition is distinct from an invalid query argument and from corrupted serialized data.

The C++ and Java implementations preserve this distinction with exception categories:

- DataSketches C++ T-Digest and KLL throw `std::runtime_error` for queries on empty sketches and `std::invalid_argument` for invalid query parameters.
- DataSketches Java T-Digest throws `SketchesStateException` for queries on empty sketches and `SketchesArgumentException` for invalid query parameters.
- DataSketches Java KLL is a notable inconsistency: it reports empty-query failures as `SketchesArgumentException`.

References:

- C++ T-Digest query behavior: https://github.com/apache/datasketches-cpp/blob/master/tdigest/include/tdigest_impl.hpp
- C++ KLL query behavior: https://github.com/apache/datasketches-cpp/blob/master/kll/include/kll_sketch_impl.hpp
- Java T-Digest query behavior: https://github.com/apache/datasketches-java/blob/main/src/main/java/org/apache/datasketches/tdigest/TDigestDouble.java
- Java KLL query behavior: https://github.com/apache/datasketches-java/blob/main/src/main/java/org/apache/datasketches/kll/KllDoublesSketch.java

## Problem

The crate does not yet have a documented policy for these independent conditions:

1. The sketch is valid but contains no data.
2. The caller supplies an invalid rank, NaN value, or invalid split points.
3. Serialized input is malformed.
4. An internal invariant fails, indicating an implementation bug.

As a result, similar query APIs differ in return types, panic behavior, and error classification. Adding more batch query APIs will make that inconsistency more visible.

## Design questions

We should decide which Rust model applies consistently to T-Digest, KLL, and REQ queries.

### A. Keep absence separate from caller errors

```rust
fn quantile(&self, rank: f64) -> Result<Option<f64>, Error>;
```

- `Ok(Some(value))`: query succeeded.
- `Ok(None)`: the sketch is empty.
- `Err(InvalidArgument)`: the rank is invalid.

This is precise but introduces nested handling for every query.

### B. Model an undefined query as a distinct recoverable error

```rust
fn quantile(&self, rank: f64) -> Result<f64, Error>;
```

For example, extend `ErrorKind` with a distinct category such as `EmptySketch` or `UndefinedOperation`:

- `Ok(value)`: query succeeded.
- `Err(EmptySketch)`: the sketch is valid but the requested statistic is undefined.
- `Err(InvalidArgument)`: the query parameter is invalid.
- `Err(InvalidData)`: serialized input is malformed.

This is the closest Rust analogue to the C++ and Java T-Digest distinction while keeping one return layer.

### C. Treat invalid query parameters as programming errors

```rust
fn quantile(&self, rank: f64) -> Option<f64>;
```

- `Some(value)`: query succeeded.
- `None`: the sketch is empty.
- Panic: the query parameter violates the documented precondition.

This matches the current T-Digest API and is compact, but it makes runtime-provided query parameters non-recoverable without prevalidation.

### D. Make invalid inputs unrepresentable

Validated `NormalizedRank` and split-point wrapper types could leave queries returning `Option`, but the additional public types and conversion steps may be disproportionate to the problem.

The discussion should also decide:

- Whether `min_value` and `max_value` should remain `Option`, following ordinary Rust collection accessors, even if statistical query methods return `Result`.
- Whether query arguments are validated before checking emptiness, so invalid input has deterministic behavior independent of sketch state.
- Whether internal invariant failures remain assertions or `unreachable!` rather than becoming public recoverable errors.
- Whether the policy applies only to quantile sketches initially or to all sketch families.

## Desired outcome

Agree on and document a crate-wide error policy for sketch queries, then apply it consistently to the affected APIs with tests covering empty sketches, invalid arguments, malformed serialized data, and internal invariants at the appropriate boundaries.

Phoenix500526:
Distinguish valid empty sketches from invalid query arguments across T-Digest, KLL, and REQ without changing deserialization errors.

CLOSES #266

### DIFF
--- a/benchmarks/tdigest/compress.rs
+++ b/benchmarks/tdigest/compress.rs
@@ -38,7 +38,7 @@ fn initial_buffer(bencher: Bencher) {
 fn unmerged_tail(bencher: Bencher) {
     let values = values(3_280);
     let mut digest = build_mut_digest(&values[..1_640]);
-    black_box(digest.rank(0.5));
+    black_box(digest.rank(0.5).unwrap().unwrap());
     for &value in &values[1_640..] {
         digest.update(value);
     }
--- a/benchmarks/tdigest/merge.rs
+++ b/benchmarks/tdigest/merge.rs
@@ -34,8 +34,8 @@ fn merge(bencher: Bencher) {
     let values = values(200_000);
     let mut left = build_mut_digest(&values[..100_000]);
     let mut right = build_mut_digest(&values[100_000..]);
-    black_box(left.rank(0.5));
-    black_box(right.rank(0.5));
+    black_box(left.rank(0.5).unwrap().unwrap());
+    black_box(right.rank(0.5).unwrap().unwrap());
 
     bencher
         .counter(ItemsCount::new(values.len()))
@@ -65,7 +65,7 @@ fn small_partials(bencher: Bencher) {
     let partials = partial_digests(64, SMALL_ROWS_PER_PARTIAL)
         .into_iter()
         .map(|mut digest| {
-            black_box(digest.rank(0.0));
+            black_box(digest.rank(0.0).unwrap().unwrap());
             digest
         })
         .collect::<Vec<_>>();
@@ -86,7 +86,7 @@ fn partials(bencher: Bencher) {
     let partials = partial_digests_with(DEFAULT_DIGEST_K, 64, ROWS_PER_PARTIAL)
         .into_iter()
         .map(|mut digest| {
-            black_box(digest.rank(0.0));
+            black_box(digest.rank(0.0).unwrap().unwrap());
             digest
         })
         .collect::<Vec<_>>();
--- a/datasketches/src/kll/mod.rs
+++ b/datasketches/src/kll/mod.rs
@@ -36,7 +36,10 @@
 //! let mut sketch = KllSketch::<i64>::new(200).unwrap();
 //! sketch.update(1);
 //! sketch.update(2);
-//! let q = sketch.quantile(0.5, SearchCriteria::Inclusive).unwrap();
+//! let q = sketch
+//!     .quantile(0.5, SearchCriteria::Inclusive)
+//!     .unwrap()
+//!     .unwrap();
 //! assert!((1..=2).contains(&q));
 //! ```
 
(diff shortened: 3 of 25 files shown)


tisonkun:
> Distinguish valid empty sketches from invalid query arguments across T-Digest, KLL, and REQ without changing deserialization errors.

Could you give a brief on the design? Otherwise I have to reverse engineering the code to understand the intention. Even if with AI assistant the intention can drift.

Phoenix500526:
> > Distinguish valid empty sketches from invalid query arguments across T-Digest, KLL, and REQ without changing deserialization errors.
> 
> Could you give a brief on the design? Otherwise I have to reverse engineering the code to understand the intention. Even if with AI assistant the intention can drift.

The query APIs currently handle empty sketches and invalid arguments inconsistently: KLL and REQ report an empty sketch as InvalidArgument, while T-Digest returns None for an empty sketch but panics on invalid query arguments. This PR adopts one contract across the three implementations. A fallible query returns Ok(Some(value)) for a valid query on a non-empty sketch, Ok(None) for a valid query on an empty sketch, and Err(InvalidArgument) for an invalid rank, NaN value, or invalid split points. Validation runs before the empty-state check, so an invalid query against an empty sketch still returns an error. Queries without an invalid-argument path, such as rank on KLL and REQ, return Option directly. Deserialization continues to return InvalidData; the serialization format and deserialization behavior are unchanged.

The implementation applies this contract to mutable and immutable T-Digest queries and to both Sketch and SortedView for KLL and REQ. The rest of the diff updates callers, examples, benchmarks, and existing tests to the new return types, with regression coverage for empty queries, invalid arguments, and their precedence.

PR #279, opened 2026-09-16
fix(tdigest): preserve min/max when merging digests

### LINKED ISSUE #277: T-Digest: merge() silently corrupts min/max
I asked Claude to look for bugs and it reported this against the [T-digest merge operation](https://github.com/apache/datasketches-rust/blob/main/datasketches/src/tdigest/sketch.rs#L337-L345)

>  T-Digest merge() silently corrupts min/max — HIGH
datasketches/src/tdigest/sketch.rs:337-345
// merge() never reads other.min / other.max —
// it infers them from compress_sorted_centroids(), which takes
// the extreme centroids' *means*, not their true min/max
T-Digest already stores explicit min/max fields — a correct merge just needs self.min = self.min.min(other.min). Instead it re-derives them from centroid means, which is only exact when the extreme centroids have unit weight. After ordinary compression this need not hold. Result: min_value()/max_value() drift, and rank()'s tail-interpolation formula silently degrades. This is a pure merge-algorithm defect — the input digest doesn't need to be malformed in any way, just legitimately compressed.

Does this seem like a valid bug?

jaideeppyne:
## Summary

- Update `TDigestMut::merge` to take `self.min`/`self.max` from `other.min`/`other.max` before compression.
- Compression only sees centroid means, which can differ from the stored extrema (e.g. after deserialization with heavy extreme centroids), so merge previously drifted `min_value()`/`max_value()` and degraded tail rank/quantile interpolation.
- Add a regression that merges digests whose stored min/max differ from extreme centroid means.
- Document the fix in the changelog.

Fixes #277

## Note for reviewers

This branch also touches `.github/workflows/ci.yml` only because the fork was behind `main` and the OAuth token lacks the `workflow` scope needed to push a tip that restores the current upstream workflow file. Please treat that hunk as unrelated fork-sync noise; the intended change is the three non-workflow files. Happy to refresh the branch once `workflow` scope is available.

## Validation

- Regression constructed from the existing `deserialize_with_centroids` helper (empty digest merge + two-way merge with drifted extrema).
- Diff against `main` for the t-digest sources is the two-line min/max update plus the new test and changelog line.


jaideeppyne:
cc @tisonkun — would appreciate a review when you have a moment. Core fix is the two-line min/max update in `merge`; please disregard the unrelated `.github/workflows/ci.yml` hunk (fork sync / missing `workflow` OAuth scope).

tisonkun:
@jaideeppyne You should be able to update fork repo's main branch so that you don't bring back old CI hacks

jaideeppyne:
Updated my fork main to upstream main (fast-forward) and removed the obsolete CI workflow changes from this PR. Its diff now contains only the TDigest fix, regression test, and changelog. I also corrected the test’s endpoint-rank expectations to match TDigest half-sample semantics. Local cargo x check, cargo x test, and cargo x lint pass.

tisonkun (approved):
Thanks!

PR #280, opened 2026-09-16
fix(tdigest): preserve stored min and max when merging

### LINKED ISSUE #277: T-Digest: merge() silently corrupts min/max

jaideeppyne:
Fixes #277.

## Summary

- Copy the stored `min` / `max` from the other digest in `TDigestMut::merge` before compressing the combined centroid list.
- Add a regression that merges a deserialized image whose extreme centroids have weight > 1 (so their means are not the true extrema).
- Document the bug fix in the changelog.

## Why

Compression never merges the extreme centroids for digests built through `update`, so this is invisible on that path. Reference-format and other deserialized images can store min/max strictly outside the extreme centroid means. `merge` previously ignored those fields and re-derived extrema from centroid means, so:

- `min_value()` / `max_value()` drifted inward
- rank / quantile tail interpolation used the wrong endpoints

This is the same class of tail-interpolation defect as #250, on the merge path rather than the query formulas.

## Validation

- `cargo test -p tests-integration --test tdigest_test`
- `cargo +nightly clippy -p datasketches --all-features -- -D warnings`


jaideeppyne:
Closing as duplicate of #279 (same #277 fix).

